PR #7, opened 2018-12-10
HTBHF-92 Modified the initial db script, including renaming it. Updated the readme.

YetAnotherMatt:


db/src/main/resources/db/migration/V1_001__create_claimant_table.sql
```diff
@@ -0,0 +1,4 @@
+create table claimant (
+    id uuid not null primary key,
+    claimant_name varchar(100) not null
```
mikeSmith08: As this is in the `claimant` table, should this field be just name then, otherwise all our fields might turn out to be `claimant_*` ?

YetAnotherMatt: Done

PR #15, opened 2018-12-20
 HTBHFB-114 Adding Flyway to perform db migration on startup

YetAnotherMatt:


YetAnotherMatt:
Don't merge this yet! We need to change the db script

PR #45, opened 2019-01-28
HTBHF-423 Added address entity to the claimant.

dwybourn:


db/src/main/resources/db/migration/V1_007__add_address.sql
```diff
@@ -0,0 +1,10 @@
+create table address (
+    id uuid not null primary key,
+    address_line_1 varchar(500) not null,
+    address_line_2 varchar(500),
+    town_or_city varchar(500) not null,
+    postcode varchar(7) not null
+);
+
+alter table claimant add column address_id uuid;
```
YetAnotherMatt: might want to be specific about what address this is = e.g. card_delivery_address_id

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
```diff
@@ -0,0 +1,77 @@
...
+    @Column(name = "address_line_1")
+    private String addressLine1;
+
+    @Size(max = 500)
+    @Column(name = "address_line_2")
+    private String addressLine2;
+
+    @NotNull
+    @Size(min = 1, max = 500)
+    @Column(name = "townOrCity")
```
mikeSmith08: Column name doesn't match the format of other column names (underscores)

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class AddressTest {
+
+    // Create a string 501 characters long
+    private static final String LONG_STRING = CharBuffer.allocate(501).toString().replace('\0', 'A');
+
+    private Validator validator;
+
+    @BeforeEach
+    void setup() {
```
mikeSmith08: Might it be worthwhile having a `AbstractValidationTest` and move this into there?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
```diff
@@ -0,0 +1,112 @@
...
+        //Given
+        var address = anAddressWithAddressLine1(LONG_STRING);
+        //When
+        Set<ConstraintViolation<Address>> violations = validator.validate(address);
+        //Then
+        assertThat(violations).hasSingleConstraintViolation("size must be between 1 and 500", "addressLine1");
+    }
+
+    @Test
+    void shouldFailToValidateAddressWhenAddressLine2IsTooLong() {
```
mikeSmith08: I'd also add a test to make sure that a null addressLine2 is valid.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
```diff
@@ -0,0 +1,112 @@
...
+        //Given
+        var address = anAddressWithPostcode(null);
+        //When
+        Set<ConstraintViolation<Address>> violations = validator.validate(address);
+        //Then
+        assertThat(violations).hasSingleConstraintViolation("must not be null", "postcode");
+    }
+
+    @Test
+    void shouldFailToValidateAddressWithInvalidPostcode() {
```
mikeSmith08: Can we have a parameterised test with lots of values here please rather than just a single one? Blank String, too long, too short, invalid characters etc etc.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/AddressTestDataFactory.java
```diff
@@ -0,0 +1,69 @@
+package uk.gov.dhsc.htbhf.claimant.testsupport;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import uk.gov.dhsc.htbhf.claimant.entity.Address;
+
+/**
+ * Test data factory for {@link uk.gov.dhsc.htbhf.claimant.entity.Address} objects.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
```
mikeSmith08: Pointless??

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
mikeSmith08: In fact the column name doesn't match, we should have a failing unit test for this...

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claimant.java
```diff
@@ -52,6 +55,9 @@
     @Column(name = "date_of_birth")
     private LocalDate dateOfBirth;
 
+    @OneToOne(mappedBy = "claimant", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
```
mikeSmith08: Do we want this to be lazy loaded, wouldn't we normally want the address available straight away too? I'm not sure, its just a question on whether we should load it at the same time as the Claimant we'd normally need it.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
dwybourn: I've updated the name but uncertain as to why the repository test passed. Potentially some hibernate magic or h2 mismatch? Once I've added the api layer in I'll see if this breaks the application.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
```diff
@@ -0,0 +1,109 @@
...
+        //Given
+        var address = anAddressWithPostcode(null);
+        //When
+        Set<ConstraintViolation<Address>> violations = validator.validate(address);
+        //Then
+        assertThat(violations).hasSingleConstraintViolation("must not be null", "postcode");
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"AA1122BB", "A", "11AA21", "", "E!", "EA123"})
```
mikeSmith08: Can we have a blank String in here too please?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/AddressTestDataFactory.java
mikeSmith08: I really do think this is pointless by the way...

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
dwybourn: There is, fourth argument.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/AddressTestDataFactory.java
dwybourn: We can remove it as a pmd rule. @YetAnotherMatt are you ok with that?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
```diff
@@ -0,0 +1,82 @@
...
+@Data
+@Builder
+@Table(name = "address")
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class Address {
+
+    /**
+     * Regex for matching UK postcodes matching BS7666 format.
+     * { @see https://www.gov.uk/government/publications/bulk-data-transfer-for-sponsors-xml-schema }
```
YetAnotherMatt: Is it worth mentioning that the file with the regex definition is BulkDataCommon-v2.1.xsd?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
```diff
@@ -0,0 +1,82 @@
...
+     * { @see https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive }
+     */
+    public static final String UK_POST_CODE_REGEX = "([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})"
+            + "|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\\s?[0-9][A-Za-z]{2})";
+    @Id
+    @Getter(AccessLevel.NONE)
+    @Access(AccessType.PROPERTY)
```
YetAnotherMatt: Why have we got `@Access(AccessType.PROPERTY)`? I think we should be using the getter instead

api/src/main/resources/application.yml
```diff
@@ -23,3 +23,5 @@ spring:
       default-property-inclusion: NON_NULL
     flyway:
         baseline-on-migrate: true
+server:
+    port: 8090
```
YetAnotherMatt: do we need this? We only need it to be 8090 when running locally, don't we?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/AddressTestDataFactory.java
YetAnotherMatt: Yes, lets remove the PMD rule in test

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
```diff
@@ -0,0 +1,82 @@
...
+public class Address {
+
+    /**
+     * Regex for matching UK postcodes matching BS7666 format.
+     * { @see https://www.gov.uk/government/publications/bulk-data-transfer-for-sponsors-xml-schema }
+     * { @see https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive }
+     */
+    public static final String UK_POST_CODE_REGEX = "([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})"
+            + "|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\\s?[0-9][A-Za-z]{2})";
+    @Id
```
YetAnotherMatt: if we're editing this anyway, blank line between the constant and @Id :)

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Address.java
dwybourn: Added reference.

api/src/main/resources/application.yml
dwybourn: Removed accidental checkin

PR #50, opened 2019-01-30
Feature/expected due date

YetAnotherMatt:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+        ClaimDTO claimDTO = ClaimDTOTestDataFactory.aValidClaimDTO();
+
+        // When
+        controller.newClaim(claimDTO);
+
+        // Then
+        Iterable<Claimant> claimants = claimantRepository.findAll();
+        assertThat(claimants).hasSize(1);
+        Claimant persistedClaim = claimants.iterator().next();
+        AssertionsForClassTypes.assertThat(persistedClaim.getNino()).isEqualTo(claimDTO.getClaimant().getNino());
```
mikeSmith08: What has happened to all these `assertThat`s in this test, is there any reason for them to use `AssertionsForClassTypes?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/constraint/DateWithinRelativeRange.java
```diff
@@ -0,0 +1,88 @@
...
+            return true;
+        }
+        return isWithinValidRange(value);
+    }
+
+    private boolean isWithinValidRange(LocalDate value) {
+        LocalDate lastValidDate = LocalDate.now().plusMonths(monthsInFuture);
+        boolean isBeforeEnd = value.isBefore(lastValidDate) || value.isEqual(lastValidDate);
+
+        LocalDate firstValidDate = LocalDate.now().minusMonths(monthsInPast);
```
mikeSmith08: Although very unlikely, it is possible that the date provided by this `LocalDate.now()` call is different to the one above so I think we should have a variable at the start of this method. Also, as the date is constructed in this class, this can make testing dates tricky (although not too bad in this scenario). Should we consider maybe having a component or similar to provide the date/time so that we can control the date for testing? Possibly something similar to having a Clock in the application context that we can mock?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+public class ClaimControllerIntegrationTest {
+
+    @Autowired
+    ClaimController controller;
+
+    @Autowired
+    ClaimantRepository claimantRepository;
+
+    @Test
+    void shouldPersistNewClaimant() {
```
mikeSmith08: Would probably be clearer called `shouldPersistNewClaimantWithoutExpectedDeliveryDate`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
```diff
@@ -144,4 +144,40 @@ void shouldFailToValidateClaimantWithDateOfBirthInFuture() {
...
+    @Test
+    void shouldValidateClaimantWithExpectedDueDateInThePast() {
+        //Given
+        Claimant claimant = ClaimantTestDataFactory.aValidClaimantBuilder()
+                .expectedDeliveryDate(LocalDate.now().minusYears(3))
+                .build();
+        //When
+        Set<ConstraintViolation<Claimant>> violations = validator.validate(claimant);
+        //Then
+        assertThat(violations).hasNoViolations();
```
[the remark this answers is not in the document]
YetAnotherMatt: As discussed, I've added a comment to Claimant explaining why

PR #51, opened 2019-01-30
Feature/card delivery address

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/converter/ClaimDTOToClaimConverter.java
```diff
@@ -33,7 +34,13 @@ private Claimant buildClaimant(ClaimantDTO claimant) {
...
                 .nino(claimant.getNino())
-                .cardDeliveryAddress(addressConverter.convert(claimant.getCardDeliveryAddress()))
+                .cardDeliveryAddress(convertCardDeliveryAddress(claimant))
                 .build();
     }
+
+    private Address convertCardDeliveryAddress(ClaimantDTO claimant) {
+        return claimant.getCardDeliveryAddress() == null
+                ? null
+                : addressConverter.convert(claimant.getCardDeliveryAddress());
```
YetAnotherMatt: As discussed, if this were staying long-term I'd rather it was a full if statement, but it'll do short-term

PR #52, opened 2019-01-30
Feature/make card address mandatory

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claimant.java
```diff
@@ -58,6 +58,7 @@
     // no validation here as we don't want to prevent re-saving a claimant 6 months after initially created
     private LocalDate expectedDeliveryDate;
 
+    @NotNull
```
YetAnotherMatt: We need to consider what happens to the existing claims in the db

YetAnotherMatt: i.e. if they're updated at any point. It may simply be that we decide to delete all existing data and not worry about it.

PR #56, opened 2019-02-12
HTBHF-535 added a session id (from X-SESSION-ID request header) to log messages

YetAnotherMatt:


api/src/main/resources/logback.xml
```diff
@@ -2,7 +2,7 @@
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} [%thread] [%X{request.id}] %-5level %logger{36} - %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] [%X{session.id}][%X{request.id}] %-5level %logger{36} - %msg%n</pattern>
```
mikeSmith08: Isn't this going to make log lines very log with two UUIDs (plus date, thread id etc) in them before we even get to the log message?

PR #63, opened 2019-03-18
Feature/call eligibility service

mikeSmith08:
Add in the call to the eligibility service. We have left this such that we still need to add the status on the claimant in the database, plus any failure in the call to the eligibility service has no impact on the saving of the Claimant so that we can add it in without it breaking the CD build (hopefully).

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/PersonDTO.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/PersonDTO.java
@@ -1,0 +1,34 @@
+package uk.gov.dhsc.htbhf.claimant.model.eligibility;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.dhsc.htbhf.claimant.model.AddressDTO;
+
+import java.time.LocalDate;
+
+/**
+ * Representation of what is required to  call the Eligibility Service.
+ */
+@Data
+@Builder
+@AllArgsConstructor(onConstructor_ = {@JsonCreator})
+public class PersonDTO {
+
+    @JsonProperty("firstName")
+    private final String firstName;
+
+    @JsonProperty("lastName")
+    private final String lastName;
+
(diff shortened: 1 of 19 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/PersonDTO.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.time.LocalDate;
+
+/**
+ * Representation of what is required to  call the Eligibility Service.
+ */
+@Data
+@Builder
+@AllArgsConstructor(onConstructor_ = {@JsonCreator})
+public class PersonDTO {
```
mikeSmith08: We specifically didn't add validation annotations to this class as its never validated - we might want to do the same thing to the `DWPPersonDTO` in the eligibility service.

api/src/main/resources/application.yml
```diff
@@ -30,4 +30,10 @@ logging:
...
         ROOT: ${vcap.services.variable-service.credentials.claimant-root-loglevel:info}
-        uk.gov.dhsc: ${vcap.services.variable-service.credentials.claimant-app-loglevel:debug}
\ No newline at end of file
+        uk.gov.dhsc: ${vcap.services.variable-service.credentials.claimant-app-loglevel:debug}
+
+eligibility:
+    base-uri: ${ELIGIBILITY_URI:http://localhost:8100}
+
+server:
+    port: 8090
```
dwybourn: spring:boot:admin needs updating as well.

manifest.yml
```diff
@@ -10,7 +10,8 @@ applications:
   env:
     JBP_CONFIG_OPEN_JDK_JRE: '{jre: { version: 11.+ }}'
     APP_VERSION: ((app-version))
+    ELIGIBILITY_URI: http://htbhf-eligibility-service((app-suffix)).apps.internal:8080
```
YetAnotherMatt: This needs to be space-suffix, not app-suffix. app-suffix includes '-green'

api/src/main/resources/application.yml
mikeSmith08: Good spot, done.

manifest.yml
mikeSmith08: Sorry, copy-paste error, done.

PR #65, opened 2019-03-19
HTBHF-679 Handle rest exceptions in eligibility client

gethyn1:
- Handle rest exceptions in eligibility client
- Ensure claimant is always saved, even on exceptions

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
@@ -1,9 +1,11 @@
 package uk.gov.dhsc.htbhf.claimant.service;
 
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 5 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
```diff
@@ -32,14 +35,20 @@ public EligibilityClient(@Value("${eligibility.base-uri}") String baseUri,
...
-        if (HttpStatus.OK != response.getStatusCode()) {
-            throw new EligibilityClientException(response.getStatusCode());
+        try {
+            ResponseEntity<EligibilityResponse> response = restTemplateWithIdHeaders.postForEntity(
+                    eligibilityUri,
+                    person,
+                    EligibilityResponse.class
+            );
+
+            if (HttpStatus.OK != response.getStatusCode()) {
```
YetAnotherMatt: I think we'll need object (in)equality here

PR #67, opened 2019-03-19
Feature/add status to claimant

gethyn1:
- Persist claim eligibility status to database
- Set eligibility status to ERROR if error returned from eligibility status
- Update claimant test factory with distinct methods for claimant with / without eligibility status

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/EligibilityStatus.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/EligibilityStatus.java
@@ -5,6 +5,7 @@ import lombok.AllArgsConstructor;
 
 /**
  * The possible states that a claim can be in according to the DWP.
+ * The value of ERROR has been added for when there are problems connecting to DWP.
  */
 @AllArgsConstructor
 @JsonFormat(shape = JsonFormat.Shape.STRING)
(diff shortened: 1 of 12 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/EligibilityStatus.java
```diff
@@ -5,6 +5,7 @@
 
 /**
  * The possible states that a claim can be in according to the DWP.
+ * The value of ERROR has been added for when there are problems connecting to DWP.
```
dwybourn: I'd suggest making the comment more generic, e.g. issues when talking to DWP.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimServiceTest.java
```diff
@@ -42,14 +43,20 @@ public void shouldSaveClaimant() {
...
 
         //then
-        verify(claimantRepository).save(claim.getClaimant());
+        Claimant expectedClaimant = claimant.toBuilder().eligibilityStatus(EligibilityStatus.ELIGIBLE).build();
+        verify(claimantRepository).save(expectedClaimant);
         verify(client).checkEligibility(claimant);
     }
 
     @Test
+    @SuppressWarnings("PMD.DataflowAnomalyAnalysis")
```
dwybourn: I would personally disable this rule as it's in the Controversial category and seems to cause a lot of false positives.

PR #74, opened 2019-03-27
Feature/check duplicate claims

gethyn1:
- Check database for existing claim (based on NINO)
- Return a response based on eligibility status
- Catch exceptions in claim service: save claim with ERROR status and then re-throw exception (requires suppression of PMD warning)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -12,24 +12,36 @@
...
+                eligibilityStatus = EligibilityStatus.DUPLICATE;
+            } else {
+                EligibilityResponse eligibilityResponse = client.checkEligibility(claimant);
+                eligibilityStatus = eligibilityResponse.getEligibilityStatus();
+                claimant.setHouseholdIdentifier(eligibilityResponse.getHouseholdIdentifier());
+            }
+
+            saveClaimant(claimant, eligibilityStatus);
+            return claimant;
+        } catch (Exception e) {
```
YetAnotherMatt: looks like you don't need this catch block, just put 
```
            saveClaimant(claimant, eligibilityStatus);
            return claimant;
```
in a `finally` block

YetAnotherMatt: actually, not the `return claimant` - just saveClaimant

mikeSmith08: The thing we like about this way is that you can then remove the `eligibilityStatus` field being outside the try block and default it in the catch block. So its swapping one PMD warning for another.

mikeSmith08: There are pros and cons to both from the discussions Gethyn and I are currently having.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
```diff
@@ -25,18 +37,32 @@
...
         // Then
+        ClaimResponse claimResponse = aClaimResponseWithEligibilityStatus(eligibilityStatus);
+        assertThat(response).isNotNull();
+        assertThat(response.getStatusCode()).isEqualTo(httpStatus);
+        assertThat(response.getBody()).isEqualTo(claimResponse);
+        verify(converter).convert(dto);
         verify(claimService).createClaim(convertedClaim);
     }
 
+    private static Stream<Arguments> provideStatusesForInvokeClaimService() {
```
YetAnotherMatt: I think these could be done in a csv source rather than a method. Down to personal preference, I guess.

gethyn1: This has been updated to use `@CsvSource`

PR #75, opened 2019-03-27
Fix IllegalArgumentException in compatibility tests

gethyn1:
The performance tests are creating responses from the eligibility service with a status of `NOMATCH` which need to be handled by the claimant service.

This PR ensures the claimant service handles all possible eligibility statuses.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
@@ -54,8 +54,17 @@ public class ClaimController {
 
     private ResponseEntity createResponseFromClaimant(Claimant claimant) {
         EligibilityStatus eligibilityStatus = claimant.getEligibilityStatus();
-        HttpStatus statusCode = statusMap.get(eligibilityStatus);
+        HttpStatus statusCode = getHttpStatus(eligibilityStatus);
         ClaimResponse body = ClaimResponse.builder().eligibilityStatus(eligibilityStatus).build();
         return new ResponseEntity(body, statusCode);
+    }
+
+    private HttpStatus getHttpStatus(EligibilityStatus eligibilityStatus) {
+        HttpStatus statusCode = statusMap.get(eligibilityStatus);
+        if (statusCode == null) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
```diff
@@ -34,7 +34,11 @@
     private final ClaimDTOToClaimConverter converter;
     private final Map<EligibilityStatus, HttpStatus> statusMap = Map.of(
             EligibilityStatus.ELIGIBLE, HttpStatus.CREATED,
-            EligibilityStatus.DUPLICATE, HttpStatus.OK
+            EligibilityStatus.INELIGIBLE, HttpStatus.OK,
```
YetAnotherMatt: we'll need some unit tests to confirm that we get the right status codes returned

YetAnotherMatt: ignore me, just seen the test. D'oh

api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
```diff
@@ -34,7 +34,11 @@
...
+            EligibilityStatus.PENDING, HttpStatus.OK,
+            EligibilityStatus.NOMATCH, HttpStatus.OK,
+            EligibilityStatus.DUPLICATE, HttpStatus.OK,
+            EligibilityStatus.ERROR, HttpStatus.INTERNAL_SERVER_ERROR
```
YetAnotherMatt: As discussed, I think we should also have a method like this in the controller:
    ```private HttpStatus getHttpStatus(EligibilityStatus eligibilityStatus) {
        HttpStatus statusCode = statusMap.get(eligibilityStatus);
        if (statusCode == null) {
            log.warn("eligibility status without HttpStatus: {}", eligibilityStatus);
            statusCode = HttpStatus.INTERNAL_SERVER_ERROR;
        }
        return statusCode;
    }```

YetAnotherMatt: Maybe that log.warn should be log.error

PR #81, opened 2019-04-02
HTBHF-640 Return 404 status when getting a 404 from eligibility service.

dwybourn:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityResponseTestDataFactory.java
```diff
@@ -14,4 +15,11 @@ public static EligibilityResponse anEligibilityResponse() {
                 .build();
     }
 
+    public static EligibilityResponse aNonMatchingEligibilityResponse() {
```
mikeSmith08: Can we refactor this an the other method in this Test Data Factory to use a common base method returning the builder so you can just change the eligibilityStatus?

dwybourn: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
```diff
@@ -46,6 +49,12 @@ public EligibilityResponse checkEligibility(Claimant claimant) {
                 throw new EligibilityClientException(response.getStatusCode());
             }
             return response.getBody();
+        } catch (HttpClientErrorException e) {
+            if (HttpStatus.NOT_FOUND != e.getStatusCode()) {
```
gethyn1: I would maybe swap these round as I found it a bit confusing to read initially:
```
// If status code does not equal NOT_FOUND
if (e.getStatusCode() != HttpStatus.NOT_FOUND) { ... }
```

PR #82, opened 2019-04-02
HTBHF-677 check duplicate claim for household

gethyn1:
Set eligibility status to DUPLICATE if claim already exists for household


api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimantRepository.java
```diff
@@ -5,17 +5,45 @@
...
+            + "AND claimant.eligibilityStatus = 'ELIGIBLE'")
+    Long getNumberOfMatchingEligibleClaimantsWithNino(@Param("nino") String nino);
 
-    default Boolean eligibleClaimExists(String nino) {
-        return getNumberOfMatchingEligibleClaimants(nino) != 0;
+    @Query("SELECT COUNT(claimant) "
+            + "FROM Claimant claimant "
+            + "WHERE claimant.dwpHouseholdIdentifier = :dwpHouseholdIdentifier "
+            + "AND claimant.eligibilityStatus = 'ELIGIBLE'")
+    Long getNumberOfMatchingEligibleClaimantsWithDwpHouseholdIdentifier(@Param("dwpHouseholdIdentifier") String dwpHouseholdIdentifier);
```
YetAnotherMatt: how about 'countEligibleClaimantsWithDwpHouseholdIdentifier'? We'd probably want to change all 'getNumberOf' methods if we adopt this name instead.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -41,6 +43,14 @@ public Claimant createClaim(Claim claim) {
         }
     }
 
+    private EligibilityStatus determineEligibilityStatusForHouseholdIdentifier(EligibilityResponse eligibilityResponse) {
```
YetAnotherMatt: How about 'determineEligibilityStatusForHousehold'?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimantRepository.java
```diff
@@ -5,17 +5,45 @@
...
+            + "FROM Claimant claimant "
+            + "WHERE claimant.hmrcHouseholdIdentifier = :hmrcHouseholdIdentifier "
+            + "AND claimant.eligibilityStatus = 'ELIGIBLE'")
+    Long countEligibleClaimantsWithHmrcHouseholdIdentifier(@Param("hmrcHouseholdIdentifier") String hmrcHouseholdIdentifier);
+
+    default boolean eligibleClaimExistsForNino(String nino) {
+        return countEligibleClaimantsWithNino(nino) != 0;
+    }
+
+    default boolean eligibleClaimExistsForHousehold(Optional<String> dwpHouseholdIdentifier, Optional<String> hmrcHouseholdIdentifier) {
```
dwybourn: I don't think this kind of logic belongs in the repository. I'm also not a fan of passing optionals into methods, I'd have the checks done in the calling class.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimantRepository.java
```diff
@@ -5,17 +5,45 @@
...
+        if (dwpHouseholdIdentifier.isPresent()) {
+            return countEligibleClaimantsWithDwpHouseholdIdentifier(dwpHouseholdIdentifier.get()) != 0;
+        }
+
+        if (hmrcHouseholdIdentifier.isPresent()) {
+            return countEligibleClaimantsWithHmrcHouseholdIdentifier(hmrcHouseholdIdentifier.get()) != 0;
+        }
+
```
YetAnotherMatt: I think we need to consider the scenario where a claimant moves from CTC to UC within a single cycle. At that point I think it's possible that both HMRC and DWP will return ELIGIBLE with a household identifier. The logic that's in this method might sit better in the claim service - the repository shouldn't really have decision logic.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -41,6 +43,14 @@ public Claimant createClaim(Claim claim) {
         }
     }
 
+    private EligibilityStatus determineEligibilityStatusForHousehold(EligibilityResponse eligibilityResponse) {
+        boolean eligibleClaimExistsForHousehold = claimantRepository.eligibleClaimExistsForHousehold(
+                Optional.ofNullable(eligibilityResponse.getDwpHouseholdIdentifier()),
```
dwybourn: Rather than wrap this ourselves we could make the identifiers optional in the json object.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimantRepository.java
gethyn1: This has been moved out of the repository and optionals removed

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimantRepository.java
gethyn1: Logic now in service

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -41,6 +41,28 @@ public Claimant createClaim(Claim claim) {
...
+        }
+
+        String hmrcHouseholdIdentifier = eligibilityResponse.getHmrcHouseholdIdentifier();
+
+        if (hmrcHouseholdIdentifier != null) {
+            return claimantRepository.eligibleClaimExistsForHmrcHousehold(hmrcHouseholdIdentifier);
+        }
+
+        return false;
+    }
```
YetAnotherMatt: There is a chance (e.g. if a household moves from CTC to UC within the past cycle) that we get both household identifiers back. At that point we should check against both identifiers:
```
    private boolean eligibleClaimExistsForHousehold(EligibilityResponse eligibilityResponse) {
        String dwpHouseholdIdentifier = eligibilityResponse.getDwpHouseholdIdentifier();
        boolean dwpClaimExists = dwpHouseholdIdentifier != null &&  claimantRepository.eligibleClaimExistsForDwpHousehold(dwpHouseholdIdentifier);

         String hmrcHouseholdIdentifier = eligibilityResponse.getHmrcHouseholdIdentifier();
         boolean hmrcClaimExists = hmrcHouseholdIdentifier != null && claimantRepository.eligibleClaimExistsForHmrcHousehold(hmrcHouseholdIdentifier);

         return dwpClaimExists || hmrcClaimExists;
    }
```

PR #84, opened 2019-04-03
[NO_JIRA] Fix branch on config sub-module.

mikeSmith08:
This should stop us getting the detached HEAD situation.


YetAnotherMatt:
Can we include the git config changes too?
```git config status.submodulesummary 1
git config diff.submodule log```

PR #88, opened 2019-04-09
HTBHF-896 Refactored logic of determining eligibility status into own…

dwybourn:
… class.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityStatusCalculator.java
```diff
@@ -0,0 +1,30 @@
...
+
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Service;
+import uk.gov.dhsc.htbhf.claimant.model.eligibility.EligibilityResponse;
+import uk.gov.dhsc.htbhf.claimant.repository.ClaimantRepository;
+import uk.gov.dhsc.htbhf.eligibility.model.EligibilityStatus;
+
+@Service
+@AllArgsConstructor
+public class EligibilityStatusCalculator {
```
mikeSmith08: Now this is out into its own class, I think some JavaDoc describing the rules it applies would be useful.

dwybourn: Added javadoc.

PR #90, opened 2019-04-10
Feature/add claim status

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -18,31 +22,42 @@
     private final EligibilityClient client;
     private final EligibilityStatusCalculator eligibilityStatusCalculator;
 
+    private final Map<EligibilityStatus, ClaimStatus> eligibilityStatusMapper = ImmutableMap.of(
+            EligibilityStatus.ELIGIBLE, ClaimStatus.NEW,
```
YetAnotherMatt: We're missing EligibilityStatus.INELIGIBLE -> REJECTED

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -18,31 +22,42 @@
     private final EligibilityClient client;
     private final EligibilityStatusCalculator eligibilityStatusCalculator;
 
+    private final Map<EligibilityStatus, ClaimStatus> eligibilityStatusMapper = ImmutableMap.of(
```
YetAnotherMatt: This might be better called 'statusMap'?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimServiceTest.java
```diff
@@ -36,19 +38,25 @@
     @Mock
     private EligibilityStatusCalculator eligibilityStatusCalculator;
 
-    @Test
-    public void shouldSaveNewClaimant() {
+    @ParameterizedTest(name = "Should save claimant with claim status set to {1} when eligibility status is {0}")
+    @CsvSource({
+            "ELIGIBLE, NEW",
```
YetAnotherMatt: we need INELIGIBLE here too. Can we add a test that uses an EligibilityStatus EnumSource and confirms that the persisted claimStatus is not null?

db/src/main/resources/db/migration/V1_006__add_claim_status_column.sql
```diff
@@ -0,0 +1,7 @@
+alter table claimant add column claim_status varchar(20);
+
+update claimant set claim_status = 'ACTIVE' where claim_status is null and eligibility_status = 'ELIGIBLE';
+
+update claimant set claim_status = 'REJECTED' where claim_status is null and eligibility_status != 'ELIGIBLE';
```
YetAnotherMatt: this could actually just be `update claimant set claim_status = 'REJECTED' where claim_status is null;`

db/src/main/resources/db/migration/V1_006__add_claim_status_column.sql
```diff
@@ -0,0 +1,7 @@
...
+
+alter table claimant alter column claim_status set not null;
```
YetAnotherMatt: I think we'll need an index on claim_status

dwybourn: Added index

db/src/main/resources/db/migration/V1_006__add_claim_status_column.sql
dwybourn: Updated.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimServiceTest.java
dwybourn: Added missing case.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
dwybourn: Added missing case.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
dwybourn: Renamed.

PR #93, opened 2019-04-10
Add VoucherEntitlement to Claim

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -34,33 +37,46 @@
...
-            claimant.setHmrcHouseholdIdentifier(eligibilityResponse.getHmrcHouseholdIdentifier());
-            return eligibilityStatusCalculator.determineEligibilityStatus(eligibilityResponse);
+            EligibilityStatus eligibilityStatus = eligibilityStatusCalculator.determineEligibilityStatus(eligibilityResponse);
+            return eligibilityResponse.toBuilder()
+                    .eligibilityStatus(eligibilityStatus)
+                    .build();
         }
     }
 
+    private void updateClaimFromEligibilityResponse(Claim claim, EligibilityResponse eligibilityResponse) {
```
mikeSmith08: If you have to mutate an object from a method, I always like to return the mutated object from that method, it makes it slightly clearer that you are doing so.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimServiceTest.java
```diff
@@ -78,6 +87,37 @@ public void shouldSaveNonExistingClaimant(EligibilityStatus eligibilityStatus, C
...
+        given(eligibilityStatusCalculator.determineEligibilityStatus(any())).willReturn(ELIGIBLE);
+        given(entitlementCalculator.calculateVoucherEntitlement(any(), any())).willReturn(voucherEntitlement);
+
+        //when
+        Claim result = claimService.createClaim(claimDTO);
+
+        //then
+        assertThat(result).isEqualTo(claim);
+        assertThat(result.getVoucherEntitlement()).isEqualTo(voucherEntitlement);
+        Claimant expectedClaimant = buildExpectedClaimant(claimant, ClaimStatus.NEW, ELIGIBLE);
```
mikeSmith08: Would it also be worthwhile checking that the Claimant on the result is equal to this? Otherwise if you're just using this to verify what's saved, then I'd rename it to `expectedSavedClaimant` for clarity.

PR #103, opened 2019-04-15
HTBHF-1048 Change from using pounds to pence to store voucher entitle…

mikeSmith08:
…ment amounts to avoid any potential rounding or formatting problems.


YetAnotherMatt:
> As we're now making a breaking change can we change the endpoint to v2.

The field(s) being changed are not yet used in the front-end, so while it is technically not backwards-compatible it won't break the front-end

mikeSmith08:
Yes, I'd rather not up the version number if we don't have to for this, it can be incorporated into what Gethyn is doing now before his changes go in.

PR #104, opened 2019-04-15
HTBHF-1065 Changed ClaimAudit to use NewClaimEvent subclass of event

YetAnotherMatt:
This cannot be merged until the common-logging PR is merged, and is here as an example of how the changes there could work.


YetAnotherMatt:
it's also going to fail to build until the common-logging pr is merged

api/build.gradle
```diff
@@ -38,6 +38,7 @@ group = 'uk.gov.dhsc.htbhf'
 sourceCompatibility = 1.11
 
 repositories {
+    mavenLocal()
```
YetAnotherMatt: This allows us to use local builds of common projects

api/src/main/java/uk/gov/dhsc/htbhf/claimant/ApiDocumentation.java
```diff
@@ -62,6 +68,11 @@ private ApiInfo getApiInfo() {
 
     @EventListener(ApplicationReadyEvent.class)
     public void logAfterStartup() {
-        log.info("Application started. App version={}, app id={}, instance index={}", appVersion, applicationId, instanceIndex);
+        eventLogger.logEvent(ApplicationStartedEvent.builder()
```
mikeSmith08: I know this isn't specifically done with this PR, but it feels like this method is in the wrong place, could we move it to `ClaimantServiceApplication`, I'd expect it to be somewhere more central like that rather than being associated with the Swagger doc generation.

mikeSmith08: BTW, this is looking really nice now! :-)

YetAnotherMatt: agreed, it wasn't really in the right place

PR #105, opened 2019-04-15
HTBHF-893 Updated card job to call new card service.

dwybourn:
PR is not 100% ready for merging but putting up now as most people get in before me.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/CardClientException.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/CardClientException.java
@@ -3,6 +3,9 @@ package uk.gov.dhsc.htbhf.claimant.exception;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.client.RestClientException;
 
+/**
+ * Exception used when calling the card service from the {@link uk.gov.dhsc.htbhf.claimant.service.CardClient}.
+ */
 public class CardClientException extends RuntimeException {
 
     private static final long serialVersionUID = 1L;
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/CardClientException.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/CardClientException.java
@@ -1,0 +1,17 @@
+package uk.gov.dhsc.htbhf.claimant.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.client.RestClientException;
+
+public class CardClientException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public CardClientException(HttpStatus httpStatus) {
+        super("Response code from card service was not CREATED, received: " + httpStatus.value());
+    }
+
+    public CardClientException(RestClientException restClientException, String endpointUrl) {
+        super("Exception caught trying to call card service at: " + endpointUrl, restClientException);
+    }
+}
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/CardRequestFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/CardRequestFactory.java
@@ -7,6 +7,9 @@ import uk.gov.dhsc.htbhf.claimant.entity.Claim;
 import uk.gov.dhsc.htbhf.claimant.model.AddressDTO;
 import uk.gov.dhsc.htbhf.claimant.model.card.CardRequest;
 
+/**
+ * Creates a {@link CardRequest} instance from a given {@link Claim}.
+ */
 @Service
 @AllArgsConstructor
 public class CardRequestFactory {
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/CardRequestFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/CardRequestFactory.java
@@ -1,0 +1,27 @@
+package uk.gov.dhsc.htbhf.claimant.factory;
+
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Service;
+import uk.gov.dhsc.htbhf.claimant.converter.AddressToAddressDTOConverter;
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+import uk.gov.dhsc.htbhf.claimant.model.AddressDTO;
+import uk.gov.dhsc.htbhf.claimant.model.card.CardRequest;
+
+@Service
+@AllArgsConstructor
+public class CardRequestFactory {
+
+    private AddressToAddressDTOConverter addressConverter;
+
+    public CardRequest createCardRequest(Claim claim) {
+        AddressDTO address = addressConverter.convert(claim.getClaimant().getCardDeliveryAddress());
+        return CardRequest.builder()
+                .address(address)
+                .firstName(claim.getClaimant().getFirstName())
+                .lastName(claim.getClaimant().getLastName())
+                .dateOfBirth(claim.getClaimant().getDateOfBirth())
+                .claimId(claim.getId().toString())
+                .build();
+
+    }
+}
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/SchedulerHeaderInterceptor.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/SchedulerHeaderInterceptor.java
@@ -13,7 +13,7 @@ import static uk.gov.dhsc.htbhf.requestcontext.RequestIdFilter.SESSION_ID_HEADER
 
 /**
  * Creates an http interceptor used to be used by the scheduler. The interceptor
- * creates random request and session ids and appends them to the incoming request headers.
+ * generates a request and session id and appends them to the incoming request headers.
  * This is required as {@link uk.gov.dhsc.htbhf.requestcontext.HeaderInterceptor} only works in
  * the scope of a web request.
  */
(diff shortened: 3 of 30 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
```diff
@@ -29,7 +30,7 @@
     private final String eligibilityUri;
 
     public EligibilityClient(@Value("${eligibility.base-uri}") String baseUri,
-                             RestTemplate restTemplateWithIdHeaders,
+                             @Qualifier("restTemplateWithIdHeaders") RestTemplate restTemplateWithIdHeaders,
```
dwybourn: Now we have two rest templates we need this anywhere we use a rest template.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/CardClientException.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Exception used when calling the card service from the {@link uk.gov.dhsc.htbhf.claimant.service.CardClient}.
+ */
+public class CardClientException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    public CardClientException(HttpStatus httpStatus) {
+        super("Response code from card service was not CREATED, received: " + httpStatus.value());
```
mikeSmith08: Minor point but you're mixing HttpStatus codes and names here, can we have either (the same) or both please?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/SchedulerHeaderInterceptor.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.UUID;
+
+import static uk.gov.dhsc.htbhf.requestcontext.RequestIdFilter.REQUEST_ID_HEADER;
+import static uk.gov.dhsc.htbhf.requestcontext.RequestIdFilter.SESSION_ID_HEADER;
+
+/**
+ * Creates an http interceptor used to be used by the scheduler. The interceptor
+ * generates a request and session id and appends them to the incoming request headers.
+ * This is required as {@link uk.gov.dhsc.htbhf.requestcontext.HeaderInterceptor} only works in
+ * the scope of a web request.
```
YetAnotherMatt: Not quite true, the `HeaderInterceptor` works with a `RequestContext` - which we need to be updating when we create the context for a scheduled job (so that log messages identify which scheduled job they are running). So the session id needs to be the id of the current job. We do want the request id to be different every time we make a request, I expect.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
```diff
@@ -3,23 +3,27 @@
...
 
-    /**
-     * Print out first 10 new claims (testing purposes only).
-     */
-    @Transactional
-    public void createNewCards() {
-        claimRepository.getNewClaims().limit(10).forEach(
-                claim -> log.info("Creating card for new claim {}", claim.getId())
-        );
+    public void createNewCards(List<UUID> claimIds) {
```
mikeSmith08: This should really be dealing with a single claimId at a time in a @Transactional method and not be dealing with all the claims, it should be the responsibility of the Job and batching process to deal with the numbers of claims and handing them off to the ClaimService in the correct manner.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestTemplate;
+import uk.gov.dhsc.htbhf.claimant.exception.CardClientException;
+import uk.gov.dhsc.htbhf.claimant.model.card.CardRequest;
+import uk.gov.dhsc.htbhf.claimant.model.card.CardResponse;
+
+/**
+ * Service for interacting with the card services api..
```
mikeSmith08: Minor - extra full stop, I'm surprised that Checkstyle doesn't moan at this :-).

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -0,0 +1,51 @@
...
+@Service
+@Slf4j
+public class CardClient {
+
+    private static final String CARDS_ENDPOINT = "/v1/cards";
+    private final String cardsUri;
+    private final RestTemplate restTemplate;
+
+    /**
+     * Uses a non request scope rest template due to being outside the context of a request.
```
mikeSmith08: Once you've changed how the RestTemplate works you'll want to say that this reflects uses the sessionId requestId from the job.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/CardRequestFactory.java
```diff
@@ -0,0 +1,30 @@
...
+ * Creates a {@link CardRequest} instance from a given {@link Claim}.
+ */
+@Service
+@AllArgsConstructor
+public class CardRequestFactory {
+
+    private AddressToAddressDTOConverter addressConverter;
+
+    public CardRequest createCardRequest(Claim claim) {
+        AddressDTO address = addressConverter.convert(claim.getClaimant().getCardDeliveryAddress());
```
mikeSmith08: Minor, but you should assert that the Claim isn't null here too, plus add a test for it.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
```diff
@@ -222,17 +221,16 @@ void shouldReturnLiveClaimDoesExistWhenHmrcHouseholdIdentifierExists(ClaimStatus
 
     @Test
     @Transactional
-    void shouldReturnNewClaims() {
+    void shouldReturnNewClaimsIds() {
```
mikeSmith08: `shouldReturnNewClaimIds`

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -0,0 +1,51 @@
...
+                      @Qualifier("schedulerRestTemplate") RestTemplate restTemplate) {
+        this.cardsUri = baseUri + CARDS_ENDPOINT;
+        this.restTemplate = restTemplate;
+    }
+
+    public CardResponse createNewCardRequest(CardRequest cardRequest) {
+        try {
+            ResponseEntity<CardResponse> response = restTemplate.postForEntity(cardsUri, cardRequest, CardResponse.class);
+            if (response.getStatusCode() != HttpStatus.CREATED) {
+                throw new CardClientException(response.getStatusCode());
```
YetAnotherMatt: The response body might contain useful information - we need to either log it or include it in the exception.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityClient.java
YetAnotherMatt: In which case we need to be using a constant for the qualifier - hopefully there is one in common-rest?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/NewCardServiceTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Mock
+    private ClaimRepository claimRepository;
+
+    @InjectMocks
+    private NewCardService newCardService;
+
+    @Test
+    void shouldCallCardClientForEachNewClaim() {
+        List<UUID> claimIds = List.of(UUID.randomUUID(), UUID.randomUUID());
```
mikeSmith08: I know this is very unlikely, but could we make these two UUIDs be constants from Strings so they will never clash and cause a build to fail.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/CardRequestTestDataFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class CardRequestTestDataFactory {
+
+    private static final String EMAIL = "test@email.com";
+    private static final String MOBILE = "07700900000";
+
+    public static CardRequest aValidCardRequest() {
+        return CardRequest.builder()
+                .address(aValidAddressDTO())
+                .claimId(UUID.randomUUID().toString())
```
mikeSmith08: This would be better as a constant value in case we need to assert on it.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/CardResponseTestDataFactory.java
```diff
@@ -0,0 +1,14 @@
...
+
+import uk.gov.dhsc.htbhf.claimant.model.card.CardResponse;
+
+import java.util.UUID;
+
+public class CardResponseTestDataFactory {
+
+    public static CardResponse aCardResponse() {
+        return CardResponse.builder()
+                .cardAccountId(UUID.randomUUID().toString())
```
mikeSmith08: Same here, I think a constant would be better.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/NewCardServiceTest.java
dwybourn: Seems like overkill seeing as UUID uses SecureRandom

PR #108, opened 2019-04-16
Feature/create new card

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param baseUri      base url of the card services api.
+     * @param restTemplate rest template which sets the session id to the current job id.
+     */
+    public CardClient(@Value("${card.base-uri}") String baseUri,
+                      @Qualifier(SCHEDULER_REST_TEMPLATE_QUALIFIER) RestTemplate restTemplate) {
+        this.cardsUri = baseUri + CARDS_ENDPOINT;
+        this.restTemplate = restTemplate;
+    }
+
+    public CardResponse createNewCardRequest(CardRequest cardRequest) {
```
YetAnotherMatt: this should probably be called 'requestNewCard' or 'createNewCard'

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
```diff
@@ -4,22 +4,28 @@
...
     @Transactional
-    public void createNewCards() {
-        claimRepository.getNewClaims().limit(10).forEach(
-                claim -> log.info("Creating card for new claim {}", claim.getId())
-        );
+    public void createNewCards(UUID claimId) {
+        Claim claim = claimRepository.findById(claimId)
+                .orElseThrow(() -> new EntityNotFoundException("Unable to find claim with id " + claimId));
+        CardRequest cardRequest = cardRequestFactory.createCardRequest(claim);
+        cardClient.createNewCardRequest(cardRequest);
```
YetAnotherMatt: We should add event logging here (or maybe in the method calling this), but I'm happy to leave that and do it as part of HTBHF-941

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -15,4 +17,7 @@
     public static final int TOTAL_VOUCHER_ENTITLEMENT = 4;
     public static final int TOTAL_VOUCHER_VALUE_IN_PENCE = 1240;
 
+    public static final LocalDate DATE_OF_BIRTH = LocalDate.parse("1985-12-31");
```
YetAnotherMatt: Should we prefix these with JAMES_ to indicate that they're all the same user?

dwybourn: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
dwybourn: Renamed to requestNewCard

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
dwybourn: Added TODO

dwybourn:
Up update has been made in common-rest https://github.com/DepartmentOfHealth-htbhf/htbhf-java-common/pull/23.

Once this has gone through the build will pass.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
```diff
@@ -4,22 +4,29 @@
...
-    public void createNewCards() {
-        claimRepository.getNewClaims().limit(10).forEach(
-                claim -> log.info("Creating card for new claim {}", claim.getId())
-        );
+    //TODO DJW 2019-04-16 add event logging.
+    public void createNewCards(UUID claimId) {
+        Claim claim = claimRepository.findById(claimId)
+                .orElseThrow(() -> new EntityNotFoundException("Unable to find claim with id " + claimId));
+        CardRequest cardRequest = cardRequestFactory.createCardRequest(claim);
+        cardClient.requestNewCard(cardRequest);
```
YetAnotherMatt: We're currently missing adding the cardAccountId to the claim and updating the claim status (to something like `PENDING_FIRST_PAYMENT`). It at least requires a TODO statement (probably best as a TODO rather than increasing the size of this PR).

YetAnotherMatt: We'll also have to update the set of valid claim statuses if we add this new status. Alternatively we could look for all `ACTIVE` claims with no payment records - though such a query might get expensive. We could add a sub-status? Just thinking out loud at this point :)

dwybourn: There is a separate subtask for this which isn't in progress yet.

PR #110, opened 2019-04-17
HTBHF-893 Add event logging for new card request.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/ClaimAuditor.java
```diff
@@ -27,10 +30,18 @@ public void auditNewClaim(Claim claim) {
...
         NewClaimEvent newClaimEvent = NewClaimEvent.builder()
-                .claimantId(claim.getId())
+                .claimId(claim.getId())
                 .claimStatus(claim.getClaimStatus())
                 .eligibilityStatus(claim.getEligibilityStatus())
                 .build();
         eventLogger.logEvent(newClaimEvent);
     }
+
+    public void auditNewCard(UUID claimId, CardResponse cardResponse) {
```
YetAnotherMatt: @mikeSmith08 what do you think about removing the ClaimAuditor and just using the EventLogger directly? It doesn't seem to do very much anymore.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/ClaimAuditor.java
```diff
@@ -27,10 +30,18 @@ public void auditNewClaim(Claim claim) {
...
+        NewCardEvent newCardEvent = NewCardEvent.builder()
+                .claimId(claimId)
+                .cardAccountId(cardResponse.getCardAccountId())
```
mikeSmith08: We should check for a null `CardResponse` as the new claim event logging does above.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/ClaimAuditor.java
mikeSmith08: I think it's still worthwhile as it nicely encapsulates the creating of the Event which the calling client then doesn't need to concern itself with.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/ClaimAuditor.java
dwybourn: Done.

PR #112, opened 2019-04-18
HTBHF-1163 Create message table, entity, repository and test them.

mikeSmith08:


db/src/main/resources/db/migration/V1_013__create_message_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE message_queue(
+    id uuid not null primary key,
+    message_timestamp TIMESTAMP not null DEFAULT NOW(),
+    message_type varchar(50) not null,
+    message_payload text not null
```
YetAnotherMatt: Do we know what indexes we'll want? It'll either be [message_type, message_timestamp] or [message_timestamp, message_type] depending on whether we're querying for specific message types or all messages...

mikeSmith08: Yep, sounds about right, I'll add one now.

PR #121, opened 2019-04-29
HTBHF-1162 Changed entitlement calculation to be done against a given…

dwybourn:
… date instead of the current date.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/EntitlementCalculator.java
```diff
@@ -41,41 +43,52 @@ public EntitlementCalculator(
         this.voucherValueInPence = voucherValueInPence;
     }
 
-    public VoucherEntitlement calculateVoucherEntitlement(Claimant claimant, List<LocalDate> childrenDatesOfBirth) {
-        int numberOfChildrenUnderFour = getNumberOfChildrenUnderFour(childrenDatesOfBirth);
-        int numberOfChildrenUnderOne = getNumberOfChildrenUnderOne(childrenDatesOfBirth);
+    /**
+     * Calculates the {@link VoucherEntitlement} a claimant is entitled to for a given date. The number of children under one and
```
mikeSmith08: This JavaDoc isn't completely accurate anymore as its given the DoB of all children.

dwybourn: It is, but the comment is still correct. We are calculating the number of children under one and four.

PR #124, opened 2019-05-01
Feature/remove pregnancy vouchers

gethyn1:
Update voucher entitlement when new children are matched to expected delivery date during payment cycle.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -20,11 +20,25 @@
...
+     * Constructor for {@link CycleEntitlementCalculator}.
+     * @param paymentCycleDurationInDays duration of payment cycle
+     * @param numberOfCalculationPeriods number of calculation periods in cycle
+     * @param weeksBeforeDueDate number of weeks before expected due date that matches a new child to that expected due date
+     * @param weeksAfterDueDate number of weeks after expected due date that matches a new child to that expected due date
+     * @param entitlementCalculator calculates entitlement for a single calculation period
+     */
     public CycleEntitlementCalculator(@Value("${payment-cycle.cycle-duration-in-days}") Integer paymentCycleDurationInDays,
                                       @Value("${payment-cycle.number-of-calculation-periods}") Integer numberOfCalculationPeriods,
+                                      @Value("${payment-cycle.child-matched-to-pregnancy-period.weeks-before-due-date}") Integer weeksBeforeDueDate,
```
mikeSmith08: I do think it would be worthwhile encapsulating these 4 config parameters into a bean of their own for simplification, plus we don't have 4 Integers floating around which could be misused/mistaken.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -43,15 +57,53 @@ private void validateArguments(Integer paymentCycleDurationInDays, Integer numbe
...
         List<VoucherEntitlement> voucherEntitlements = entitlementDates.stream()
-                .map(date -> entitlementCalculator.calculateVoucherEntitlement(claimant, dateOfBirthOfChildren, date))
+                .map(date -> entitlementCalculator.calculateVoucherEntitlement(expectedDueDate, dateOfBirthOfChildren, date))
                 .collect(Collectors.toList());
 
         return new PaymentCycleVoucherEntitlement(voucherEntitlements);
     }
 
+    /**
+     * Calculates the total voucher entitlement for a payment cycle.
```
mikeSmith08: How about `Calculates the total voucher entitlement for a payment cycle given the previous entitlement` so it's clear what the difference is from the other method.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -43,15 +57,53 @@ private void validateArguments(Integer paymentCycleDurationInDays, Integer numbe
...
+     * @param expectedDueDate expected due date
+     * @param dateOfBirthOfChildren the date of birth of the claimant's children
+     * @param previousVoucherEntitlement voucher entitlement from last payment cycle
+     * @return the payment cycle voucher entitlement calculated for the claimant
+     */
+    public PaymentCycleVoucherEntitlement calculateEntitlement(Optional<LocalDate> expectedDueDate,
+                                                               List<LocalDate> dateOfBirthOfChildren,
+                                                               PaymentCycleVoucherEntitlement previousVoucherEntitlement) {
+        if (newChildrenMatchedToExpectedDeliveryDate(expectedDueDate, dateOfBirthOfChildren, previousVoucherEntitlement)) {
+            // ignore expected due date as we've determined that the pregnancy has happened
```
mikeSmith08: Don't you mean `birth has happened` or `pregnancy has finished`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -43,15 +57,53 @@ private void validateArguments(Integer paymentCycleDurationInDays, Integer numbe
...
+        }
+
+        return calculateEntitlement(expectedDueDate, dateOfBirthOfChildren);
+    }
+
+    // A child is considered a result of expected due date if their date of birth falls within a range of that date
+    private boolean newChildrenMatchedToExpectedDeliveryDate(Optional<LocalDate> expectedDueDate,
+                                                             List<LocalDate> dateOfBirthOfChildren,
+                                                             PaymentCycleVoucherEntitlement previousVoucherEntitlement) {
+        if (previousVoucherEntitlement.getVouchersForPregnancy() == 0 || expectedDueDate.isEmpty()) {
```
mikeSmith08: How about `if(noPreviousPregnancyVouchers() || notPregnant())`, so let's refactor those two out so it reads nicely.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -43,15 +57,53 @@ private void validateArguments(Integer paymentCycleDurationInDays, Integer numbe
...
+            return false;
+        }
+
+        LocalDate startDate = expectedDueDate.get().minusWeeks(weeksBeforeDueDate);
```
mikeSmith08: How about we create a method to encapsulate the period matching, so this becomes:
```
        return dateOfBirthOfChildren.stream()
                .anyMatch(date -> isWithPregnancyMatchPeriod(expectedDueDate, date));
    }

    private boolean isWithPregnancyMatchPeriod(Optional<LocalDate> expectedDueDate, LocalDate date) {
        LocalDate startDate = expectedDueDate.get().minusWeeks(weeksBeforeDueDate);
        LocalDate endDate = expectedDueDate.get().plusWeeks(weeksAfterDueDate);
        return !date.isBefore(startDate) && !date.isAfter(endDate);
    }
```

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: The four integers could still be misused/mistaken, just in a different place. I'm not sure moving the integers to a separate bean really adds anything. Especially as these values are only used in this class.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: Updated

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: Updated to pregnancy has finished.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: done.

mikeSmith08: Sorry, my mistake here, the method should be called `isWithinPregnancyMatchPeriod`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
mikeSmith08: I still do think its worthwhile having a `PaymentCycleConfig` bean to nicely encapsulate these fields, plus you can also have the calculation of the `entitlementCalculationDuration` field within it, which means that you only expose that value and not the other two config values - I think its definitely worthwhile!!

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: I should have noticed, updated.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
dwybourn: Ok, moved values into configuration bean.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/CycleEntitlementCalculator.java
```diff
@@ -20,38 +19,82 @@
...
+     * @param entitlementCalculator calculates entitlement for a single calculation period
+     */
+    public CycleEntitlementCalculator(PaymentCycleConfig paymentCycleConfig,
                                       EntitlementCalculator entitlementCalculator) {
-        validateArguments(paymentCycleDurationInDays, numberOfCalculationPeriods);
-        this.entitlementCalculationDuration = paymentCycleDurationInDays / numberOfCalculationPeriods;
-        this.numberOfCalculationPeriods = numberOfCalculationPeriods;
+        this.weeksBeforeDueDate = paymentCycleConfig.getWeeksBeforeDueDate();
+        this.weeksAfterDueDate = paymentCycleConfig.getWeeksAfterDueDate();
+        this.entitlementCalculationDuration = paymentCycleConfig.getEntitlementCalculationDuration();
```
gethyn1: Should we just store `paymentCycleConfig` instead of defining each individual integer? Then just call them when required.

dwybourn: I think in the code referencing weeksBeforeDueDate looks nice that paymentCycleConfig.getWeeksBeforeDueDate().

PR #129, opened 2019-05-08
HTBHF-1238 Modify pregnancy calculator to use same period as payment …

mikeSmith08:
…cycle. Add integration tests for checking that vouchers stop when no birth is notified.

### DIFF
--- a/api/src/main/resources/application.yml
+++ b/api/src/main/resources/application.yml
@@ -41,7 +41,6 @@ entitlement:
   number-of-vouchers-per-child-under-one: 2
   number-of-vouchers-per-child-between-one-and-four: 1
   number-of-vouchers-per-pregnancy: 1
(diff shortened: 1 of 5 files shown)


api/src/main/resources/application.yml
```diff
@@ -41,7 +41,6 @@ entitlement:
   number-of-vouchers-per-child-under-one: 2
   number-of-vouchers-per-child-between-one-and-four: 1
   number-of-vouchers-per-pregnancy: 1
-  pregnancy-grace-period-in-days: 42
```
YetAnotherMatt: Where is pregnancy grace period configured now?

YetAnotherMatt: Never mind - found it.

YetAnotherMatt: Never mind - found it.

PR #136, opened 2019-05-10
Feature/add payment cycle

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/PaymentCycleService.java
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
+import uk.gov.dhsc.htbhf.claimant.repository.PaymentCycleRepository;
+
+@Service
+@AllArgsConstructor
+public class PaymentCycleService {
+    private PaymentCycleRepository paymentCycleRepository;
+
+    public void createNewPaymentCycle(Claim claim) {
```
YetAnotherMatt: This will need to accept a cycleStartDate and populate the cycleEndDate as well, but that can be in a subsequent commit.

dwybourn: We're doing that work now for a future PR.

PR #137, opened 2019-05-10
HTBHF-1286 Created a query to identify payment cycles ending today

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleEndDate.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.UUID;
+
+/**
+ * Identifies a claim with a payment cycle due to expire on cycleEndDate.
+ */
+public interface PaymentCycleEndDate {
+
+    String getClaimIdString();
+
+    Timestamp getCycleEndDateTimestamp();
```
mikeSmith08: Why a `Timestamp` and not a `LocalDateTime`? This differs from all the other date/time constructs we are using.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleEndDate.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.dhsc.htbhf.claimant.repository;
+
+import java.sql.Timestamp;
+import java.time.LocalDate;
+import java.util.UUID;
+
+/**
+ * Identifies a claim with a payment cycle due to expire on cycleEndDate.
+ */
+public interface PaymentCycleEndDate {
```
mikeSmith08: Feels like this should have a better name as it refers to the end date itself rather than the PaymentCycle, so it should be something like `PaymentCycleWithEndDate` or with an appropriate adjective like `DefinedPaymentCycle`, `ClosedPaymentCycle`, `BoundedPaymentCycle` or `ActivePaymentCycle`. I've listed a few there as I'm not too happy with most of those, but you get the idea. It's slightly harder as it isn't used yet.

YetAnotherMatt: How about ExpiringPaymentCycle?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleEndDate.java
YetAnotherMatt: This is the result of experimentation. Initially I had [UUID, LocalDate], but the native query returns a String and a Timestamp - hence the default methods to convert them to the desired values.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleEndDate.java
YetAnotherMatt: or ClosingPaymentCycle perhaps. Looking for more synonyms of 'ending'...

YetAnotherMatt: I reckon either ClosingPaymentCycle or PaymentCycleWithEndDate.

mikeSmith08: Naming is always hard isn't it! How about `ClosingPaymentCycle`, then should we need another field on here it won't seem wrong unlike `PaymentCycleWithEndDate`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleEndDate.java
mikeSmith08: I see, fine, the calling code should be using `getCycleEndDate` and not `getCycleEndDateTimestamp`, ok, that's fine.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleRepositoryTest.java
```diff
@@ -0,0 +1,163 @@
...
+    @Autowired
+    private PaymentRepository paymentRepository;
+
+    @AfterEach
+    void afterEach() {
+        paymentCycleRepository.deleteAll();
+        paymentRepository.deleteAll();
+        claimRepository.deleteAll();
+    }
+
```
dwybourn: Could we have an EnumSource test that checks with claims for each status.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleRepository.java
```diff
@@ -1,12 +1,31 @@
...
  */
 public interface PaymentCycleRepository extends CrudRepository<PaymentCycle, UUID> {
+
+    /**
+     * Looks for all active (ACTIVE, PENDING_EXPIRY) claims whose latest PaymentCycle has an end date on or before cycleEndDate.
+     * @param cycleEndDate the date by which the cycle should end.
+     * @return a List of {@link ClosingPaymentCycle}s containing the claim id and cycle end date.
+     */
+    @Query(nativeQuery = true, value =
+            "SELECT Cast(claim_id as varchar) as claimIdString, cycle_end_date as cycleEndDateTimestamp FROM "
```
dwybourn: We should create a plan to stress test this at some point if we're selecting all. See how many entries it takes to cause memory or performance issues.

YetAnotherMatt: I do plan to test it locally with my 500k active claims

YetAnotherMatt: We will also need another environment (Pre-Prod) in which we can load a lot of claims and leave it running.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentCycleRepositoryTest.java
YetAnotherMatt: Can't do an EnumSource as different statuses produce different results. I have done 2 CsvSources - one with claims that should be included in the results, another with claims that shouldn't.

YetAnotherMatt:
After the merge this has become a much smaller commit :)

PR #139, opened 2019-05-13
Feature/make payment message

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleService.java
```diff
@@ -0,0 +1,32 @@
...
+    private final PaymentCycleRepository paymentCycleRepository;
+    private final Integer cycleDurationInDays;
+
+    public PaymentCycleService(PaymentCycleRepository paymentCycleRepository,
+                               @Value("${payment-cycle.cycle-duration-in-days}") Integer cycleDurationInDays) {
+        this.paymentCycleRepository = paymentCycleRepository;
+        this.cycleDurationInDays = cycleDurationInDays;
+    }
+
+    public void createNewPaymentCycle(Claim claim) {
```
YetAnotherMatt: Creating new paymentCycles for existing claims will require passing a start date parameter rather than using the claimStatusTimestamp. Can that be done in the calling method instead and add cycleStartDate as a parameter?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleService.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private final PaymentCycleRepository paymentCycleRepository;
+    private final Integer cycleDurationInDays;
+
+    public PaymentCycleService(PaymentCycleRepository paymentCycleRepository,
+                               @Value("${payment-cycle.cycle-duration-in-days}") Integer cycleDurationInDays) {
+        this.paymentCycleRepository = paymentCycleRepository;
+        this.cycleDurationInDays = cycleDurationInDays;
+    }
+
```
dwybourn: github diff is misleading here. This isn't a new file, it's been modified and moved to a new package. Same goes for the test.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentService.java
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.dhsc.htbhf.claimant.message.MessageQueueDAO;
+import uk.gov.dhsc.htbhf.claimant.message.MessageType;
+import uk.gov.dhsc.htbhf.claimant.message.payload.MessagePayload;
+
+@Service
+@AllArgsConstructor
+@Slf4j
+public class PaymentService {
+
+    private MessageQueueDAO messageQueueDAO;
```
mikeSmith08: I think this should be injected by the interface `MessageQueueClient` then when/if we decide to swap implementation to RabbitMQ, this code will not need to change. Also, the code to use it below looks a bit odd as we're calling `sendMessage` on a DAO.

dwybourn: Updated to use the interface.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleService.java
dwybourn: Updated to take in cycle start date.

PR #140, opened 2019-05-13
HTBHF-1287 Created a ClaimLazyLoader to create Claims with a given ID…

YetAnotherMatt:
… without returning to the db

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimLazyLoader.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimLazyLoader.java
@@ -1,0 +1,19 @@
+package uk.gov.dhsc.htbhf.claimant.repository;
+
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+
+import java.util.UUID;
+
+/**
+ * Clients should not refer to this interface directly - use ClaimRepository instead.
+ */
+public interface ClaimLazyLoader {
+
+    /**
+     * Creates a reference to a Claim object without loading that Claim from the database.
+     * @param id the id of the Claim to create a reference for.
+     * @return An uninitialised proxy for a Claim object.
+     */
+    Claim getLazyLoadingClaim(UUID id);
+}
+
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
@@ -244,14 +244,4 @@ class ClaimRepositoryTest {
         //Then
         assertThat(result).containsOnly(newClaim.getId());
     }
-
-    @Test
-    void shouldReturnLazyLoadedReferenceToClaim() {
-        UUID id = UUID.randomUUID();
-
-        Claim result = claimRepository.getLazyLoadingClaim(id);
-
-        assertThat(result).isNotNull();
-        assertThat(result.getId()).isEqualTo(id);
-    }
 }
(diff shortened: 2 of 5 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
```diff
@@ -244,4 +244,14 @@ void shouldReturnNewClaimIds() {
         //Then
         assertThat(result).containsOnly(newClaim.getId());
     }
+
+    @Test
+    void shouldReturnLazyLoadedReferenceToClaim() {
+        UUID id = UUID.randomUUID();
```
YetAnotherMatt: I'm testing here that the claim has not been loaded from the DB, as doing so would throw an error. Open to better ways of testing this.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimLazyLoader.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.dhsc.htbhf.claimant.repository;
+
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+
+import java.util.UUID;
+
+/**
+ * Clients should not refer to this interface directly - use ClaimRepository instead.
+ */
+public interface ClaimLazyLoader {
```
mikeSmith08: Can you explain the reasoning behind needing this now? Just wanted some further background before approving...

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
mikeSmith08: I'm not sure I understand this test, surely it shouldn't return a result at all as the db is empty?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimLazyLoader.java
YetAnotherMatt: It will be needed for my subtask - by the upcoming `CreateNewPaymentCycleJob`, which will need to get a reference to the claim to pass to the `PaymentService`. 
It is admittedly coming in the wrong order as it's not needed by the next PR but the one after. My next PR will be the scheduler and an empty implementation of `CreateNewPaymentCycleJob`, followed by implementing the job - which will need this.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
YetAnotherMatt: The point is that it's not going to the DB, instead creating a proxy for us with the given ID. If we were to try and access any other properties I expect we'd get an error as it would then try to call the DB to populate the Claim. 
Maybe I should just remove this test - I put it here largely to reassure myself that Spring was wiring everything up correctly.

mikeSmith08: Wow, I didn't realise that's what it was trying to achieve. Yes, this either needs re-writing so that is clear, or we just remove it.

dwybourn: Perhaps a comment above the test explaining it's purpose might help.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimLazyLoader.java
mikeSmith08: Ok, that's fine thing, just didn't have any context in this PR.

PR #147, opened 2019-05-14
HTBHF-1289 Refactored MessageContextLoader to be more generic

YetAnotherMatt:
I.e. to create MakePaymentMessageContexts as well.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
@@ -1,23 +1,23 @@
-package uk.gov.dhsc.htbhf.claimant.message;
+package uk.gov.dhsc.htbhf.claimant.message.context;
 
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 import uk.gov.dhsc.htbhf.claimant.entity.Claim;
 import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
+import uk.gov.dhsc.htbhf.claimant.message.MessageProcessingException;
 import uk.gov.dhsc.htbhf.claimant.message.payload.DetermineEntitlementMessagePayload;
+import uk.gov.dhsc.htbhf.claimant.message.payload.MakePaymentMessagePayload;
 import uk.gov.dhsc.htbhf.claimant.repository.ClaimRepository;
 import uk.gov.dhsc.htbhf.claimant.repository.PaymentCycleRepository;
 
 import java.util.Optional;
 import java.util.UUID;
 
-import static uk.gov.dhsc.htbhf.claimant.message.MessageType.DETERMINE_ENTITLEMENT;
-
 @Component
 @AllArgsConstructor
 @Slf4j
-public class DetermineEntitlementMessageContextLoader {
+public class MessageContextLoader {
 
     private ClaimRepository claimRepository;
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
```diff
@@ -42,25 +42,31 @@ public DetermineEntitlementMessageContext loadContext(DetermineEntitlementMessag
...
+
+    private Claim getAndCheckClaim(UUID claimId) {
         Optional<Claim> claim = claimRepository.findById(claimId);
         if (claim.isEmpty()) {
             logAndThrowException("claim", claimId);
         }
         return claim.get();
     }
 
-    private PaymentCycle getAndCheckCurrentPaymentCycle(DetermineEntitlementMessagePayload payload) {
```
YetAnotherMatt: I removed these methods as to do this for all message types would result in quite a lot of methods

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
```diff
@@ -70,7 +76,7 @@ private PaymentCycle getAndCheckPaymentCycle(UUID paymentCycleId, String cycleNa
     }
 
     private void logAndThrowException(String fieldName, UUID uuid) {
-        String errorMessage = String.format("Unable to process %s message, unable to load %s using id: %s", DETERMINE_ENTITLEMENT, fieldName, uuid);
+        String errorMessage = String.format("Unable to process message, unable to load %s using id: %s", fieldName, uuid);
```
YetAnotherMatt: Had to make this more generic in order to share it across message types

mikeSmith08: I think it would still be useful to pass the message type in, you can do this from the `loadContext` methods as they know which message type that are for.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoaderTest.java
```diff
@@ -18,28 +20,88 @@
...
+
+    @Test
+    void shouldFailToLoadPaymentContextIfClaimNotFound() {
+        //Given
+        UUID claimId = UUID.randomUUID();
+        PaymentCycle paymentCycle = aValidPaymentCycle();
+        UUID paymentCycleId = paymentCycle.getId();
+        when(paymentCycleRepository.findById(paymentCycleId)).thenReturn(Optional.of(paymentCycle));
+        given(claimRepository.findById(any())).willReturn(Optional.empty());
+        when(paymentCycleRepository.findById(paymentCycleId)).thenReturn(Optional.of(paymentCycle));
```
mikeSmith08: This line is duplicated in this test.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoaderTest.java
```diff
@@ -18,28 +20,88 @@
...
     @Test
-    void shouldSuccessfullyLoadContext() {
+    void shouldSuccessfullyLoadPaymentContext() {
+        //Given
+        Claim claim = ClaimTestDataFactory.aValidClaim();
+        UUID claimId = claim.getId();
+        PaymentCycle paymentCycle = aValidPaymentCycle();
+        UUID paymentCycleId = paymentCycle.getId();
+        given(claimRepository.findById(any())).willReturn(Optional.of(claim));
+        when(paymentCycleRepository.findById(paymentCycleId)).thenReturn(Optional.of(paymentCycle));
```
mikeSmith08: Any reason why you're using `when` here rather than `given`? Same in test below too.

PR #148, opened 2019-05-15
HTBHF-1289 Added the depositFunds call to CardClient

YetAnotherMatt:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/CardClientTest.java
```diff
@@ -74,4 +77,45 @@ void shouldThrowAnExceptionWhenPostCallReturnsError() {
         assertThat(exception.getMessage()).isEqualTo("Exception caught trying to call card service at: " + baseUri + "/v1/cards");
         verify(restTemplate).postForEntity(baseUri + "/v1/cards", cardRequest, CardResponse.class);
     }
+
+    @Test
+    void shouldCallDepositFunds() {
+        DepositFundsRequest request = DepositFundsRequest.builder().build();
```
mikeSmith08: I think we should have data in the request objects, same with the tests below too. Probably needs a TestDataFactory.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/CardClientTest.java
```diff
@@ -74,4 +77,45 @@ void shouldThrowAnExceptionWhenPostCallReturnsError() {
...
+        DepositFundsResponse expectedResponse = DepositFundsResponse.builder().referenceId("response id").build();
```
mikeSmith08: Again, might be good to have this in a TestDataFactory.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -49,4 +51,21 @@ public CardResponse requestNewCard(CardRequest cardRequest) {
             throw new CardClientException(e, cardsUri);
         }
     }
+
+    public DepositFundsResponse depositFundsToCard(String cardAccountId, DepositFundsRequest depositRequest) {
+        String uri = String.format("%s/%s/deposit", cardsUri, cardAccountId);
+        try {
+            ResponseEntity<DepositFundsResponse> response = restTemplate.postForEntity(uri, depositRequest, DepositFundsResponse.class);
+            if (response.getStatusCode() != HttpStatus.OK) {
```
dwybourn: could we move this into method like checkResponse(). That way this method's main code block is just three lines that read really easily.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/CardClient.java
```diff
@@ -49,4 +51,21 @@ public CardResponse requestNewCard(CardRequest cardRequest) {
...
+                log.error("Expecting an OK response from the card service api, instead received {} with response body {}",
+                        response.getStatusCode().name(), response.getBody());
+                throw new CardClientException(response.getStatusCode());
+            }
+
+            return response.getBody();
+        } catch (RestClientException e) {
+            log.error("Exception caught trying to post to {}", uri);
```
dwybourn: Are we getting much value from adding this catch block? It won't tell us much more than what's logged in the exception handler.

PR #150, opened 2019-05-15
HTBHF-1343 Disabled the scheduler config for tests.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/SchedulerConfig.java
```diff
@@ -17,6 +18,7 @@
 @Configuration
 @EnableScheduling
 @EnableSchedulerLock(defaultLockAtMostFor = "PT60M")
+@ConditionalOnProperty(name = "enable-scheduler") // annotation has no affect if the property isn't set
```
dwybourn: This annotation has not affect if the property isn't set, so no need to set it to true in src/main/resources/application.yml

PR #158, opened 2019-05-17
HTBHF-1359 add address id column

gethyn1:
- Add address id column
- Rename card delivery address field in claimant DTO


db/src/main/resources/db/migration/V1_020__add_address_column.sql
```diff
@@ -0,0 +1,3 @@
+alter table claimant add column address_id uuid;
```
dwybourn: Why not just use rename column? Then it's a one liner

dwybourn:
Has this been tested end to end locally?

db/src/main/resources/db/migration/V1_020__add_address_column.sql
YetAnotherMatt: Because during blue/green deployment the old version of the app needs to be able to write to the old column.

PR #159, opened 2019-05-17
HTBHF-1362 enabled scheduling and increased scheduler pool size

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/PaymentCycleScheduler.java
```diff
@@ -17,6 +18,7 @@
  * This scheduler simply invokes the CreateNewPaymentCycleJob for each such claim.
  */
 @Slf4j
+@Component
```
YetAnotherMatt: The only reason the MessageProcessor didn't need `@Component` is because it is created from config by an `@Bean` annotated method.

api/src/main/resources/application.yml
```diff
@@ -69,5 +73,7 @@ payment-cycle:
     maximum-lock-time: "PT59M"
     end-date-offset-days: 0
 
+enable-scheduler: true
```
YetAnotherMatt: If this is missing no schedules are run

PR #162, opened 2019-05-17
HTBHF-1380 Updated PaymentCycleService to set total voucher counts …

YetAnotherMatt:
… when updating an existing paymentCycle

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
@@ -1,13 +1,18 @@
 package uk.gov.dhsc.htbhf.claimant.entity;
 
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 import uk.gov.dhsc.htbhf.assertions.AbstractValidationTest;
+import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleVoucherEntitlement;
 
 import java.util.Set;
 import javax.validation.ConstraintViolation;
 
 import static uk.gov.dhsc.htbhf.assertions.ConstraintViolationAssert.assertThat;
-import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleTestDataFactory.*;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleTestDataFactory.aPaymentCycleWithClaim;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleTestDataFactory.aValidPaymentCycle;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleTestDataFactory.aValidPaymentCycleBuilder;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleVoucherEntitlementTestDataFactory.aValidPaymentCycleVoucherEntitlement;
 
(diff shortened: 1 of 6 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
```diff
@@ -30,4 +35,35 @@ void shouldFailToValidatePaymentCycleWithNoClaim() {
...
+    @Test
+    void shouldApplyVoucherEntitlement() {
+        //Given
+        PaymentCycle paymentCycle = aValidPaymentCycleBuilder()
+                .voucherEntitlement(null)
+                .totalEntitlementAmountInPence(null)
+                .totalVouchers(null)
+                .build();
+        PaymentCycleVoucherEntitlement entitlement = aValidPaymentCycleVoucherEntitlement();
+        Assertions.assertThat(entitlement.getTotalVoucherEntitlement()).isGreaterThan(0);
```
dwybourn: Can we use a static import here

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
```diff
@@ -30,4 +35,35 @@ void shouldFailToValidatePaymentCycleWithNoClaim() {
...
+        Assertions.assertThat(entitlement.getTotalVoucherValueInPence()).isGreaterThan(0);
```
dwybourn: Feels very weird to have assertions in the given stage. Instead could we just create an entitlement in a way in which we're explicit about the values we care about.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
YetAnotherMatt: Sadly not. There's already a static import for a different assertThat.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
YetAnotherMatt: This might be affected by another PR I just created, after which `aValidPaymentCycleVoucherEntitlement` has been renamed `aPaymentCycleVoucherEntitlementWithVouchers` so we can get rid of these asserts.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
dwybourn: Should we switch the assertJ to be a static import seeing as it's used more than the constraint violation one?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
mikeSmith08: @dwybourn we can't do that here because those values are set/calculated on construction so we are just making sure the values are as expected before we start the test.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycleTest.java
YetAnotherMatt: this is the only test (of 3) that uses this version of assertThat - the constraint violation is actually more common in this class), I think

PR #163, opened 2019-05-18
HTBHF-1379 Refactored EligibilityService to correct eligibility, return EligibilityAndEntitlement

YetAnotherMatt:
Sorry it's a big PR, but about 8 of the files are just due to a method rename.
The functional change here is that when we get an eligibility response stating that the claimant is eligible, we still have to ensure that they would get vouchers (i.e. that they are either pregnant or have young children). This is not something the DWP or HMRC can do for us, as they don't know about pregnancy.

I refactored so that this logic was performed by the EligibilityService, which means that the EligibilityService now also provides information about entitlement - returning an EligibilityAndEntitlement object. This has a side-effect of no longer passing a DTO (EligibilityResponse) around large parts of the system.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -66,48 +59,26 @@ public MessageStatus processMessage(Message message) {
...
-                dateOfBirthOfChildren,
-                currentPaymentCycle.getCycleStartDate(),
-                previousPaymentCycle.getVoucherEntitlement());
+        return COMPLETED;
     }
 
     private void updateAndSaveCurrentPaymentCycle(PaymentCycle currentPaymentCycle,
-                                                  EligibilityResponse eligibilityResponse,
-                                                  PaymentCycleVoucherEntitlement paymentCycleVoucherEntitlement) {
+                                                  EligibilityAndEntitlement eligibility) {
```
mikeSmith08: I think this should be renamed to `eligibilityAndEntitlement` as it looks really odd below when you call `eligibility.getVoucherEntitlement()`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/EligibilityAndEntitlement.java
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleVoucherEntitlement;
+import uk.gov.dhsc.htbhf.eligibility.model.EligibilityStatus;
+
+import java.time.LocalDate;
+import java.util.List;
+
+@Data
+@Builder
+@AllArgsConstructor
+public class EligibilityAndEntitlement {
```
mikeSmith08: I think there may be a better name here, something to do with the `PaymentCycle`, how about `PaymentCycleCalculation`, `PaymentCycleDetermination` or `PaymentCycleEvaluation` or something similar, I never like a class name with `And` in it.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
```diff
@@ -16,33 +23,72 @@
...
+                .eligibilityStatus(eligibilityStatus)
+                .voucherEntitlement(entitlement)
+                .dateOfBirthOfChildren(eligibilityResponse.getDateOfBirthOfChildren())
+                .dwpHouseholdIdentifier(eligibilityResponse.getDwpHouseholdIdentifier())
+                .hmrcHouseholdIdentifier(eligibilityResponse.getHmrcHouseholdIdentifier())
+                .build();
     }
+
+    private EligibilityStatus decideEligibilityStatus(EligibilityResponse response, PaymentCycleVoucherEntitlement voucherEntitlement) {
+        if (response.getEligibilityStatus() == EligibilityStatus.ELIGIBLE && voucherEntitlement.getTotalVoucherEntitlement() == 0) {
```
mikeSmith08: Is this a safe way to do this, shouldn't we be checking based on the number of children and the presence of an expected delivery date, or are we presuming that the voucher entitlement process will have correctly done its job?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
```diff
@@ -16,63 +25,140 @@
...
+        EligibilityAndEntitlement eligibilityResponse = eligibilityService.determineEligibilityForNewClaimant(claimant);
 
         assertThat(eligibilityResponse.getEligibilityStatus()).isEqualTo(DUPLICATE);
         verify(claimRepository).liveClaimExistsForNino(claimant.getNino());
-        verifyZeroInteractions(client, eligibilityStatusCalculator);
+        verifyZeroInteractions(client, eligibilityStatusCalculator, cycleEntitlementCalculator);
+    }
+
+    @Test
+    void shouldReturnEligibleWhenNoDuplicateAndEligibleWithVouchers() {
```
mikeSmith08: `NotDuplicate`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
```diff
@@ -16,63 +25,140 @@
...
+        EligibilityResponse eligibilityResponse = anEligibilityResponse();
+        PaymentCycleVoucherEntitlement voucherEntitlement = aPaymentCycleVoucherEntitlementWithVouchers();
+        given(claimRepository.liveClaimExistsForNino(any())).willReturn(false);
+        given(client.checkEligibility(any())).willReturn(eligibilityResponse);
+        given(eligibilityStatusCalculator.determineEligibilityStatusForNewClaim(any())).willReturn(ELIGIBLE);
+        given(cycleEntitlementCalculator.calculateEntitlement(any(), any(), any())).willReturn(voucherEntitlement);
+
+        EligibilityAndEntitlement result = eligibilityService.determineEligibilityForNewClaimant(claimant);
+
+        assertCorrectResponse(result, eligibilityResponse.getEligibilityStatus(), eligibilityResponse, voucherEntitlement);
```
mikeSmith08: I think you should be specific about checking the status here by passing in `ELIGIBLE`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
```diff
@@ -16,63 +25,140 @@
...
+        PaymentCycleVoucherEntitlement voucherEntitlement = aPaymentCycleVoucherEntitlementWithVouchers();
+        PaymentCycle previousCycle = PaymentCycleTestDataFactory.aValidPaymentCycle();
         given(client.checkEligibility(any())).willReturn(eligibilityResponse);
+        given(cycleEntitlementCalculator.calculateEntitlement(any(), any(), any(), any())).willReturn(voucherEntitlement);
 
-        EligibilityResponse result = eligibilityService.determineEligibilityForExistingClaimant(claimant);
+        EligibilityAndEntitlement result = eligibilityService.determineEligibilityForExistingClaimant(claimant, cycleStartDate, previousCycle);
 
-        assertThat(result).isEqualTo(eligibilityResponse);
+        assertCorrectResponse(result, eligibilityResponse.getEligibilityStatus(), eligibilityResponse, voucherEntitlement);
```
mikeSmith08: Same here, should pass in the specific eligibility status.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
```diff
@@ -16,63 +25,140 @@
...
+                eligibilityResponse.getDateOfBirthOfChildren(),
+                cycleStartDate,
+                previousCycle.getVoucherEntitlement());
         verifyZeroInteractions(eligibilityStatusCalculator, claimRepository);
     }
 
+    @Test
+    void shouldReturnIneligibleForExistingClaimantWithoutVouchers() {
+        Claimant claimant = aValidClaimant();
+        EligibilityResponse eligibilityResponse = anEligibilityResponse();
```
mikeSmith08: Can we be specific about the `EligibilityStatus` here as its hard to read without specifically setting it here, and its important to the test that we know what it is.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
YetAnotherMatt: We are indeed assuming that the entitlement calculator will have done its job correctly

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
YetAnotherMatt: I was thinking 'when there is no duplicate' - either works

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
```diff
@@ -16,33 +23,72 @@
...
      * the external eligibility service is called.
+     * Claimants determined to be eligible must still either be pregnant or have children under 4,
+     * otherwise they will be ineligible.
      *
      * @param claimant the claimant to check the eligibility for
-     * @return the eligibility response for the claimant
+     * @return the eligibility and entitlement for the claimant
      */
-    public EligibilityResponse determineEligibilityForNewClaimant(Claimant claimant) {
+    public EligibilityAndEntitlement determineEligibilityForNewClaimant(Claimant claimant) {
```
dwybourn: Should this not be renamed to determineEligibilityAndEntitlementForNewClaimant?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
YetAnotherMatt: and have 5 parameters? How about `assertCorrectEligibleResponse`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityServiceTest.java
YetAnotherMatt: I did think about renaming `anEligiblityResponse` to `anEligibleResponse` - would that work? (It would affect a lot more classes though)

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityAndEntitlementTestDataFactory.java
```diff
@@ -0,0 +1,45 @@
...
+                .eligibilityStatus(ELIGIBLE)
+                .dwpHouseholdIdentifier(DWP_HOUSEHOLD_IDENTIFIER)
+                .hmrcHouseholdIdentifier(HMRC_HOUSEHOLD_IDENTIFIER)
+                .dateOfBirthOfChildren(children);
+    }
+
+    private static List<LocalDate> createChildren(Integer numberOfChildrenUnderOne, Integer numberOfChildrenUnderFour) {
+        List<LocalDate> childrenUnderOne = nCopies(numberOfChildrenUnderOne, MAGGIE_DOB);
+        List<LocalDate> childrenBetweenOneAndFour = nCopies(numberOfChildrenUnderFour - numberOfChildrenUnderOne, LISA_DOB);
+        return Stream.concat(childrenUnderOne.stream(), childrenBetweenOneAndFour.stream()).collect(Collectors.toList());
```
mikeSmith08: This feels like hard work, we can use either Guava's `concat` method or Apache Commons `union` methods to do this without building these streams.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/VoucherEntitlementTestDataFactory.java
```diff
@@ -16,6 +16,16 @@ public static VoucherEntitlement aValidVoucherEntitlement() {
                 .build();
     }
 
+    public static VoucherEntitlement aVoucherEntitlementWithZeroVouchers() {
+        return VoucherEntitlement.builder()
+                .voucherValueInPence(VOUCHER_VALUE_IN_PENCE)
```
mikeSmith08: I keep mistaking this field for the total voucher value (I thought that its value should have been zero and took me a while to figure out why it wasn't the case). Might it be worthwhile to rename this to something like `individualVouchValueInPence`? Obviously not for this PR though...

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -59,26 +59,26 @@ public MessageStatus processMessage(Message message) {
...
                 claimant,
                 currentPaymentCycle.getCycleStartDate(),
                 previousPaymentCycle);
 
         //TODO HTBHF-1296 - update ClaimStatus from ACTIVE to PENDING_EXPIRY if Claimant is no longer eligible.
-        updateAndSaveCurrentPaymentCycle(currentPaymentCycle, eligibility);
+        updateAndSaveCurrentPaymentCycle(currentPaymentCycle, decision);
 
-        if (eligibility.getEligibilityStatus() == ELIGIBLE) {
+        if (decision.getEligibilityStatus() == ELIGIBLE) {
```
mikeSmith08: Minor - is it worthwhile adding an `isEligibile()` method to `EligibilityAndEntitlementDecision`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
```diff
@@ -34,7 +34,7 @@
      * @param claimant the claimant to check the eligibility for
      * @return the eligibility and entitlement for the claimant
      */
-    public EligibilityAndEntitlement determineEligibilityForNewClaimant(Claimant claimant) {
+    public EligibilityAndEntitlementDecision determineEligibilityAndEntitlementForNewClaimant(Claimant claimant) {
```
mikeSmith08: ...this does now make we wonder whether the name of the service (and this method) should be reviewed as the `EligibilityService` is also determining entitlement...sorry... Would this method be something like `decideForNewClaimant` or `evaluateNewClaimant`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
mikeSmith08: Yes, but my question is more whether we should when we have all the information we need without depending upon another component to have done its job, hence removing the (albeit indirect) linkage between the two components.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
YetAnotherMatt: `EligibilityAndEntitlementService`?
`.evaluateNewClaimant` & `.evaluateExistingClaimant`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityAndEntitlementTestDataFactory.java
```diff
@@ -1,35 +1,35 @@
...
+import static com.google.common.collect.Iterables.concat;
+import static com.google.common.collect.Lists.newArrayList;
 import static java.util.Collections.nCopies;
 import static uk.gov.dhsc.htbhf.claimant.testsupport.TestConstants.*;
 import static uk.gov.dhsc.htbhf.eligibility.model.EligibilityStatus.ELIGIBLE;
 
 public class EligibilityAndEntitlementTestDataFactory {
 
-    public static EligibilityAndEntitlement anEligibilityAndEntitlementWithStatus(EligibilityStatus eligibilityStatus) {
+    public static EligibilityAndEntitlementDecision anEligibilityAndEntitlementDecisionWithStatus(EligibilityStatus eligibilityStatus) {
```
mikeSmith08: Maybe just `aDecisionWithStatus` and `aDecisionWithStatusAndEntitlement` below?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
YetAnotherMatt: This is the only place it would be used, so I might leave it for now.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityService.java
gethyn1: Sounds good to me

PR #168, opened 2019-05-20
Feature/add message count

mikeSmith08:
Update message delivery count and status when processing is not successful, or a status of not COMPLETED is returned from a message type processor.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Message.java
```diff
@@ -30,4 +31,9 @@
     @Column(name = "message_payload")
     private String messagePayload;
 
+    @Column(name = "delivery_count")
+    private int deliveryCount;
+
+    @Column(name = "status")
```
dwybourn: Need an enumerated annotation.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageProcessor.java
```diff
@@ -75,18 +78,33 @@ private MessageStatus processMessage(Message message, MessageTypeProcessor messa
...
-            // TODO: HTBHF-1285: handle other message statuses
+
             return status;
         } catch (RuntimeException e) {
             // TODO: HTBHF-1285: expose ErrorHandler.constructExceptionDetail
             //                  from java common and include it in error handling
             log.error("Unable to process message with id {}", message.getId(), e);
-            // TODO: HTBHF-1285: increment delivery count and message timestamp
-            return MessageStatus.ERROR;
+            MessageStatus errorStatus = ERROR;
```
dwybourn: We can replace these two lines with updateMessagesToErrorAndIncrementCount

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Message.java
mikeSmith08: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageProcessor.java
mikeSmith08: That method can validly take different statuses so I'll leave it like that. I've also added ERROR as a variable as this is what's returned from this method.

PR #169, opened 2019-05-20
HTBHF-1388 Modified creation and update of PaymentCycle to include dates of birth

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleService.java
```diff
@@ -67,13 +73,13 @@ public PaymentCycle createAndSavePaymentCycleForEligibleClaim(Claim claim, Local
...
-     * @param paymentCycle      The payment cycle
-     * @param eligibilityStatus The eligibility status
-     * @param entitlement       The entitlement
+     * @param paymentCycle The payment cycle
+     * @param decision     The decision containing details of the claimants eligibility and entitlement during this payment cycle
      */
-    public void updateAndSavePaymentCycle(PaymentCycle paymentCycle, EligibilityStatus eligibilityStatus, PaymentCycleVoucherEntitlement entitlement) {
-        paymentCycle.setEligibilityStatus(eligibilityStatus);
-        paymentCycle.applyVoucherEntitlement(entitlement);
+    public void updateAndSavePaymentCycle(PaymentCycle paymentCycle, EligibilityAndEntitlementDecision decision) {
```
YetAnotherMatt: Should we have this method at all? My understanding of hibernate is that if we modify the PaymentCycle within a transaction the changes will be persisted - so we can just modify the PaymentCycle itself. Even if we do need to call the save method (e.g. if the cycle has been disconnected) I don't think we should have a method to which we need to keep adding parameters every time there's a change.

mikeSmith08: I quite like this method, it makes it explicit what we're doing and at what point rather than waiting for Hibernate to do at magically at some point in time.

mikeSmith08:
Would be good to know why we need to store this on the PaymentCycle, as far as I can tell, nothing in the code comments or even the JIRA ticket linked says why we now need it.

YetAnotherMatt:
> Would be good to know why we need to store this on the PaymentCycle, as far as I can tell, nothing in the code comments or even the JIRA ticket linked says why we now need it.

It's for MI & helpdesk use

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewClaimService.java
```diff
@@ -67,8 +67,8 @@ private Claim createAndSaveClaim(Claimant claimant, EligibilityAndEntitlementDec
         return claim;
     }
 
-    private void sendNewCardMessage(Claim claim, PaymentCycleVoucherEntitlement voucherEntitlement) {
-        NewCardRequestMessagePayload payload = buildNewCardMessagePayload(claim, voucherEntitlement);
+    private void sendNewCardMessage(Claim claim, EligibilityAndEntitlementDecision decision) {
+        NewCardRequestMessagePayload payload = buildNewCardMessagePayload(claim, decision.getVoucherEntitlement(), decision.getDateOfBirthOfChildren());
         messageQueueDAO.sendMessage(payload, CREATE_NEW_CARD);
```
gethyn1: This is the third variation of children's DOB list: `childrenDob`, `dateOfBirthOfChildren` and `datesOfBirthOfChildren` - should we decide on a standard?

YetAnotherMatt: `childrenDob` is persisted in PaymentCycle so changing that is a bit more cumbersome. but changing `dateOfBirthOfChildren` to `datesOfBirthOfChildren` is a good plan.

PR #170, opened 2019-05-20
Feature/add claim auditing

dwybourn:


db/README.md
```diff
@@ -109,3 +109,19 @@ Set the preferred maintenance window (For production this is currently every Sun
 ```
 cf update-service htbhf-claimant-service-postgres -c '{"preferred_maintenance_window": "Sun:03:00-Sun:03:30"}'
 ```
+
+Claim auditing
+-----------
+All changes to claims are audited using the [javer](https://javers.org/) library.
```
YetAnotherMatt: javers (with an s)

db/README.md
```diff
@@ -109,3 +109,19 @@ Set the preferred maintenance window (For production this is currently every Sun
...
 ```
+
+Claim auditing
+-----------
+All changes to claims are audited using the [javer](https://javers.org/) library. 
+
+* jv_global_id — domain object identifiers,
+* jv_commit — Javers commits metadata,
+* jv_commit_property — commit properties,
+* jv_snapshot — domain object snapshots
```
YetAnotherMatt: Don't we need a sql script to create these tables?

dwybourn: It's done automatically and seems to be the recommended option. The tables are created if they don't exist. We can change this to create them with flyway if preferred.

db/README.md
dwybourn: Fixed.

db/README.md
YetAnotherMatt: I think we should be using flyway, rather than relying on tables being created at startup

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
```diff
@@ -244,4 +250,21 @@ void shouldReturnNewClaimIds() {
...
+        claimRepository.save(claim);
+
+        //When
+        JqlQuery jqlQuery = QueryBuilder.byClass(Claim.class).build();
+        Changes changes = javers.findChanges(jqlQuery);
+
+        //Then
+        assertThat(changes.size()).isEqualTo(1);
+        assertThat(changes.get(0).toString()).isEqualTo("ValueChange{ 'claimStatus' value changed from 'ACTIVE' to 'PENDING' }");
+    }
```
YetAnotherMatt: What happend is an address is modified, or a claim is updated without calling ClaimRepository.save?

dwybourn: We don't have an address repository to update an address on it's own and we don't have any other way of updating a claim. The claim repository is the own way of updating a claim, so just added the javers annotation there.

db/README.md
dwybourn: Added flyway script.

mikeSmith08:
We seem to be running ahead with Javers here - I'm not against it, but for something which is going to give us a vital piece of functionality, I'd like to know your reasons for choosing this solution. I thought the idea was to have a Spike first before jumping into the code, so let's take a breath for a moment and go through the Spike output please.

PR #171, opened 2019-05-21
HTBHF-1390 A query limits to message retrieval repository method.

mikeSmith08:


api/src/main/resources/application.yml
```diff
@@ -57,6 +57,7 @@ message-processor:
   # for an explanation of the ISO-8601 duration format
   minimum-lock-time: "PT1M"
   maximum-lock-time: "PT60M"
+  message-limit: 100
```
YetAnotherMatt: I think we may want to make this a bit bigger than 100. Shall we try 1000 and see how fast we can process them? We probably want about as many as we can process in a minute.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageProcessor.java
```diff
@@ -44,8 +48,8 @@ public void processAllMessages() {
     }
 
     private void processMessagesOfType(MessageType messageType) {
-        List<Message> messages = messageRepository.findAllMessagesByTypeOrderedByDate(messageType);
-        if (CollectionUtils.isEmpty(messages)) {
+        List<Message> messages = messageRepository.findAllMessagesByTypeOrderedByDate(messageType, PageRequest.of(0, messageProcessingLimit));
+        if (messages.isEmpty()) {
```
dwybourn: Do we no longer have to deal with a null response?

api/src/main/resources/application.yml
mikeSmith08: Changed to 1000.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageProcessor.java
mikeSmith08: Yes, we do, I've changed it back.

PR #172, opened 2019-05-21
HTBHF-1362 Renamed voucherValueInPence to singleVoucherValueInPence

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/VoucherEntitlementDTO.java
```diff
@@ -12,21 +13,27 @@
...
+    @ApiModelProperty(notes = "The number of vouchers for pregnancy", example = "1")
     private int vouchersForPregnancy;
 
     @JsonProperty("totalVoucherEntitlement")
+    @ApiModelProperty(notes = "The total number of vouchers the claimant is entitled to", example = "4")
     private int totalVoucherEntitlement;
 
     @JsonProperty("voucherValueInPence")
+    @ApiModelProperty(notes = "The financial value of a single voucher", example = "310")
     private int voucherValueInPence;
```
YetAnotherMatt: Note that I have not changed the name used in our API - that remains voucherValueInPence, though I have added notes explaining it.

mikeSmith08: That seems an odd choice, any reason why? I think calling it `singleVoucherValueInPence` would be helpful and consistent.

YetAnotherMatt: I've added singleVoucherValueInPence - we'll need to update the front-end then we can remove voucherValueInPence.

YetAnotherMatt: Turns out it wasn't used in the front-end, so we've been able to rename it in this PR after all.

PR #175, opened 2019-05-22
HTBHF-1408 Added version number to entities that can change.

dwybourn:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
```diff
@@ -88,16 +88,6 @@ void shouldAlwaysReturnAnIdFromGetId() {
         Assertions.assertThat(id).isNotNull();
     }
 
-    @Test
-    void shouldReturnTheSameIdIfOneIsSet() {
```
dwybourn: Id is no longer available to the builder and this test isn't really needed as it was just testing a property being set. This id test wasn't done for any other entity.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/PaymentRepositoryTest.java
```diff
@@ -45,6 +45,7 @@ void shouldSavePayment() {
     @Test
     void shouldNotSaveAnInvalidPayment() {
         Payment payment = aPaymentWithCardAccountId(null);
+        claimRepository.save(payment.getClaim());
```
dwybourn: Now that claim is versioned, hibernate will throw an error about saving a payment with an unsaved claim, before throwing a validation error.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
YetAnotherMatt: as long as we have such a test for BaseEntity - it would be good to ensure that our id is always present, and remains the same.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
```diff
@@ -93,7 +94,8 @@ void shouldReturnTheSameIdIfOneIsSet() {
         //Given
         UUID id = UUID.randomUUID();
         //When
-        Claimant claimant = Claimant.builder().id(id).build();
+        Claimant claimant = Claimant.builder().build();
```
YetAnotherMatt: Does `Claimant.builder().id(id).build();` not work any more? :(

dwybourn: It doesn't unfortunately :(

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
dwybourn: I've put this test back in.

PR #177, opened 2019-05-22
HTBHF-1408 Added equality tests for entities and claim entity tests.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/VersionedEntity.java
```diff
@@ -1,14 +1,12 @@
...
 
 import lombok.*;
-import lombok.experimental.SuperBuilder;
 import org.javers.core.metamodel.annotation.DiffIgnore;
 
 import javax.persistence.*;
 
 @Data
 @NoArgsConstructor
-@SuperBuilder
```
dwybourn: Removing as it has no affect unless the child class uses a super builder. Which it's not due to no superbuilder support from the intellij plugin.

PR #180, opened 2019-05-22
HTBHF-1411 create payment calculator

gethyn1:
Create payment calculator to compare entitlement amount with card balance and determine amount for new payment request


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleVoucherEntitlement;
+
+public class PaymentCalculator {
+
+    private final int numberOfCalculationPeriods;
+    private final int calculationPeriodThreshold;
+
+    public PaymentCalculator(@Value("${payment-cycle.number-of-calculation-periods}") int numberOfCalculationPeriods) {
+        this.numberOfCalculationPeriods = numberOfCalculationPeriods;
+        this.calculationPeriodThreshold = numberOfCalculationPeriods * 2;
```
YetAnotherMatt: This needs to be configurable. I think we probably want this configured as the number of calculation periods that make up the threshold, rather than the multiplier. Naming, as ever, might be tricky. `payment-cycle.max-balance-calculation-periods`, perhaps? It'll need some description (ideally in application.yml as well as here).

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+     * @return the amount in pence to apply to the next payment request
+     */
+    public int calculatePaymentAmountCycleInPence(PaymentCycleVoucherEntitlement entitlement, int cardBalanceInPence) {
+        int totalVoucherValueInPence = entitlement.getFirstVoucherEntitlementForCycle().getTotalVoucherValueInPence();
+        int maxAllowedCardBalanceInPence = totalVoucherValueInPence * calculationPeriodThreshold;
+
+        if (cardBalanceInPence >= maxAllowedCardBalanceInPence) {
+            return 0;
+        }
+
```
YetAnotherMatt: I don't understand the purpose of partialPaymentThresholdInPence, but this is what I htink the calculation should be:
```
if (cardBalanceInPence >= maxAllowedCardBalanceInPence) {
     return 0;
}
if (cardBalanceInPence + totalVoucherValueInPence < maxAllowedCardBalanceInPence) {
    return totalVoucherValueInPence;
}
return maxAllowedCardBalanceInPence - cardBalanceInPence;
```

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculatorTest.java
```diff
@@ -0,0 +1,53 @@
...
+        int paymentAmount = paymentCalculator.calculatePaymentAmountCycleInPence(entitlement, lowBalance);
+
+        // Then
+        assertThat(paymentAmount).isEqualTo(firstWeekVoucherEntitlementAmount * 4);
+    }
+
+    @Test
+    void shouldReturnPartialPaymentAmountForBalanceBetweenFourAndEightWeeksEntitlement() {
+        // Given
+        PaymentCycleVoucherEntitlement entitlement = aPaymentCycleVoucherEntitlementWithFourWeeklyVouchers();
```
YetAnotherMatt: Can we be more explicit about what the total entitlement & max balance amounts are? I'm thinking something along the lines of (for example, not these exact values):
Given currentBalance = 20,000, maxBalance = 30,000, totalVoucherEntitlement = 25,000, paymentAmount = 10,000, or
currentBalance = 29,900, maxBalance = 30,000, totalVoucherEntitlement = 10,000, paymentAmount = 100
This might be a good candidate for a parameterised test, actually.

YetAnotherMatt: Note that in some of the above cases the totalVoucherAmount isn't 4 * weekly amount - assumption here is that one of the later weeks has a different amount. If that gets really difficult to do in the factory then don't worry about it too much.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculator.java
gethyn1: Yes, this is an inverted logic to what we had before. We have updated to follow your suggestion which I think is clearer.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculator.java
gethyn1: With help from Kate we came up with `number-of-calculation-periods`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCalculatorTest.java
gethyn1: This is quite difficult to achieve as a lot of the values are calculated internally in each voucher entitlement (which are inside a list property on PaymentCycleVoucherEntitlement). Instead we have refactored the test by adding constants to make the value assertions clearer.

PR #181, opened 2019-05-23
NO-JIRA Updated the DatabasePopulator to use multiple threads and record time taken to insert

YetAnotherMatt:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
```diff
@@ -97,19 +124,53 @@ void generateClaimsAndPaymentCycles() {
...
-                Claim claim = createClaim(status, eligibilityStatus, ninoPrefix, ninoStartIndex + i);
-                if (paymentEligibility != null) {
-                    createPaymentCycle(cycleDuration, paymentEligibility, claim);
-                }
-                if (countCreated.incrementAndGet() % 100 == 0) {
-                    log.error("Created {} claims so far", countCreated.get());
-                }
-
+                Integer ninoNumber = ninoStartIndex + i;
+                Callable<Long> job = () -> {
```
mikeSmith08: I know this is only test code, but could we extra this code to declutter this large method an make it easier to understand?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
```diff
@@ -97,19 +124,53 @@ void generateClaimsAndPaymentCycles() {
...
+                    long start = System.currentTimeMillis();
```
mikeSmith08: I like using `StopWatch` for doing these kind of time duration calculations.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
```diff
@@ -97,19 +124,53 @@ void generateClaimsAndPaymentCycles() {
...
+            for (Future<Long> future : futures) {
+                durations.add(future.get());
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        return durations;
+    }
+
+    private void logDurations(List<Long> durations) {
```
mikeSmith08: I found this quite difficult to understand, maybe it needs a refactor and some documentation to make it easier to understand?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
```diff
@@ -97,19 +117,53 @@ void generateClaimsAndPaymentCycles() {
             EligibilityStatus paymentEligibility = PAYMENT_ELIGIBILITY_STATUS.get(status);
             String ninoPrefix = NINO_PREFIX.get(status);
             Query query = entityManager.createNativeQuery("select count(*) from claimant where left(nino , 2) = '" + ninoPrefix + "'");
-            int ninoStartIndex = ((Number)query.getSingleResult()).intValue() + 1;
+            int ninoStartIndex = ((Number) query.getSingleResult()).intValue() + 10001; // ensure we have a non-zero number in the first 2 digits
             log.error("Creating {} claims in status {}, starting from {}", claims, status, ninoStartIndex);
+            List<Callable<Long>> jobs = new ArrayList<>(claims);
             for (int i = 0; i < claims; i++) {
```
dwybourn: If you create the job outside the loop, you can use the Collections.nCopies(claims, job) to create the list.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
YetAnotherMatt: I did look at doing so, but there are seven parameters we'd need to pass.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
YetAnotherMatt: I don't think I can - the variable values change each time

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
YetAnotherMatt: Extracted a NewClaimDetails object and method to create the job.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/utilities/DatabasePopulator.java
YetAnotherMatt: I added some comments, found it difficult to identify how I might refactor. Suggestions?

mikeSmith08: Having spent some time with this, this is actually quite a good way to do this, it just takes a few minutes to understand what you're doing and why. I think if you just rename a couple of the variables then this becomes a bit clearer. My thoughts are:

- Rename `map` to `durationFrequencyMap`
- Rename `bucket` to `durationRoundedToNearest10` (I know this is lengthy so feel free to rename if you can think of something better)

I did then try to refactor out the generation of the logging String as I was initially confused by the `bucket + (bucketSize -1)` part, but that was quite tricky to make it more legible so I left that alone. 

One thing I might be tempted to do was to just put an example output as a comment to that method and just say: "This method outputs a frequency map of the durations and looks like this: " and then paste in one you've done before, I think that'll do it then.

Sorry, a bit of a long comment, but hopefully that covers everything.

mikeSmith08: I should also say that I'm happy with this after that, so feel free to merge if I'm holding things up tomorrow. I would Approve but it'll only be dismissed when you do an update so no point.

PR #182, opened 2019-05-23
HTBHF-1411 Check for balance on card before making a payment and adjust payment amount accordingly.

mikeSmith08:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentService.java
```diff
@@ -53,17 +60,17 @@ public Payment makePayment(PaymentCycle paymentCycle, String cardAccountId) {
...
     }
 
-    private Integer checkBalanceAndCalculatePaymentAmount(PaymentCycle paymentCycle) {
-        // TODO: HTBHF-1267: Check balance against card provider, update paymentCycle with balance and timestamp
-        // TODO: HTBHF-1267: reduce amount paid if it would put the card over the max allowed balance
-        return paymentCycle.getTotalEntitlementAmountInPence();
+    private Integer checkBalanceAndCalculatePaymentAmount(PaymentCycle paymentCycle, String cardAccountId) {
+        CardBalanceResponse balance = cardClient.getBalance(cardAccountId);
+        paymentCycleService.updateAndSavePaymentCycleWithBalance(paymentCycle, balance);
+        return paymentCalculator.calculatePaymentAmountCycleInPence(paymentCycle.getVoucherEntitlement(), balance.getAvailableBalanceInPence());
```
gethyn1: I think we possibly made a mistake in the payment calculator. Should method name be
```
paymentCalculator.calculatePaymentCycleAmountInPence()
```
?

YetAnotherMatt: agreed :)

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentService.java
```diff
@@ -42,8 +45,12 @@ public Payment makeFirstPayment(PaymentCycle paymentCycle, String cardAccountId)
...
 
     public Payment makePayment(PaymentCycle paymentCycle, String cardAccountId) {
-        Integer amountToPay = checkBalanceAndCalculatePaymentAmount(paymentCycle);
-        // TODO: HTBHF-1267: If payment amount < minimum payment amount then don't make a payment (might need a new payment status)
+        Integer amountToPay = checkBalanceAndCalculatePaymentAmount(paymentCycle, cardAccountId);
+        if (amountToPay == 0) {
+            //TODO HTBHF-1418: Create an event for no payment made due to balance too high.
+            log.info("No payment will be made as the existing balance on the card is too high");
+            return null;
+        }
```
YetAnotherMatt: We also need to update the PaymentCycle with the amount paid (including if the amount is zero)

YetAnotherMatt: Hmmm. Just noticed that we don't have a payment amount on PaymentCycle. Leave it out of this PR then. We'll want to have a think about how we distinguish between incomplete payments and those unpaid due to card balance. I guess if the PaymentCycle has a card balance amount then it can be considered complete, and if there'sno payment it is because the balance was too high.

gethyn1: OK, I will add a subtask so we can discuss the best option next week

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentService.java
gethyn1: Updated

PR #186, opened 2019-05-28
HTBHF-1285 Added exception detail to error handling in PaymentScheduler

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageProcessor.java
```diff
@@ -86,7 +86,7 @@ private MessageStatus invokeMessageTypeProcessor(Message message, MessageTypePro
         try {
             return messageTypeProcessor.processMessage(message);
         } catch (RuntimeException e) {
-            log.error("Unable to process message with id {}, exception detail: {}", message.getId(), constructExceptionDetail(e));
+            log.error("Unable to process message with id {}, exception detail: {}", message.getId(), constructExceptionDetail(e), e);
```
YetAnotherMatt: I think we should keep the stack trace as well as the exception detail - this enables us to get the most out of local logs as well as logit.io. We have formatted the logs to exclude irrelevant frames from stack traces so it's less of a burden on logit.io than it might otherwise be.

PR #188, opened 2019-05-29
HTBHF-1425 Add status to PaymentCycle.

mikeSmith08:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycle.java
```diff
@@ -68,6 +68,10 @@
     @Column(name = "card_balance_timestamp")
     private LocalDateTime cardBalanceTimestamp;
 
+    @NotNull
+    @Column(name = "payment_cycle_status")
+    private PaymentCycleStatus paymentCycleStatus;
```
dwybourn: Needs enumerated annotation.

mikeSmith08: Done.

PR #191, opened 2019-05-29
HTBHF-1458 Claim status is set to ACTIVE when a new card is created.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
```diff
@@ -21,16 +23,21 @@
...
     public void createNewCard(Claim claim) {
         CardRequest cardRequest = cardRequestFactory.createCardRequest(claim);
         CardResponse cardResponse = cardClient.requestNewCard(cardRequest);
-        saveClaimWithCardId(claim, cardResponse);
+        saveClaimWithCardIdAndSetStatusToActive(claim, cardResponse);
         eventAuditor.auditNewCard(claim.getId(), cardResponse);
     }
 
-    private void saveClaimWithCardId(Claim claim, CardResponse cardResponse) {
+    private void saveClaimWithCardIdAndSetStatusToActive(Claim claim, CardResponse cardResponse) {
```
YetAnotherMatt: how about `updateAndSaveClaim` instead?

dwybourn: Done.

PR #197, opened 2019-05-30
HTBHF-1425 Ensure PaymentCycleStatus is set once the first payment has been completed.

htbhf-machine-user:


mikeSmith08:
Created with the wrong user.

PR #199, opened 2019-05-30
HTBHF-1449 Added updated claim event.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/EventAuditor.java
```diff
@@ -40,6 +41,19 @@ public void auditNewClaim(Claim claim) {
         eventLogger.logEvent(newClaimEvent);
     }
 
+    /**
+     * Audit an updated claim.
```
mikeSmith08: How about "Audit that a claim has been updated"?

mikeSmith08:
This event doesn't seem to be used anywhere yet, I presume that's intentional? Also, I was wondering whether it would be worthwhile having an enum with the potential field names as they have to match what the UI is expecting, so defining them somewhere would be beneficial.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/EventAuditor.java
dwybourn: Done.

PR #200, opened 2019-05-31
HTBHF-1477 Create event for card creation failed.

mikeSmith08:


mikeSmith08:
Turns out that we already have a `FailureEvent`, so I'll use that instead.

PR #205, opened 2019-05-31
HTBHF-1450 Updated dto to return updated fields and flag stating that…

dwybourn:
… the claim has been updated.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
```diff
@@ -73,7 +76,11 @@ private VoucherEntitlementDTO getEntitlement(ClaimResult result) {
         return result.getVoucherEntitlement().map(voucherConverter::convert).orElse(null);
     }
 
-    private HttpStatus getHttpStatus(ClaimStatus claimStatus) {
+    private HttpStatus getHttpStatus(ClaimResult claimResult) {
+        if (TRUE.equals(claimResult.getClaimUpdated())) {
```
mikeSmith08: I know this is correct, but it looks really ugly, can we have an `isClaimUpdated()` method on `ClaimResult` so this can just be `if(claimResult.isUpdated())`, which would look and read much better.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
```diff
@@ -103,6 +108,30 @@ void shouldAcceptAndCreateANewValidClaimWithNullFields() {
...
+        ResponseEntity<ClaimResultDTO> response = restTemplate.exchange(buildRequestEntityV2(claim), ClaimResultDTO.class);
+
+        //Then
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+        assertThat(response.getBody()).isNotNull();
+        assertThat(response.getBody().getClaimStatus()).isEqualTo(ACTIVE);
+        assertThat(response.getBody().getEligibilityStatus()).isEqualTo(ELIGIBLE);
+        assertThat(response.getBody().getVoucherEntitlement()).isEqualTo(aValidVoucherEntitlementDTO());
+        assertThat(response.getBody().getClaimUpdated()).isTrue();
+        assertThat(response.getBody().getUpdatedFields()).isEqualTo(singletonList("expectedDeliveryDate"));
```
mikeSmith08: Maybe use the enum value?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
dwybourn: As a compromise, I've updated to use isTrue from apache commons.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
dwybourn: Done.

PR #206, opened 2019-06-03
HTBHF-1450 updated UpdatableClaimFields to actually perform the update

YetAnotherMatt:
This is for comment only, NOT TO BE MERGED

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFields.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFields.java
@@ -4,104 +4,87 @@ import lombok.Getter;
 import uk.gov.dhsc.htbhf.claimant.entity.Address;
 import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.*;
-import java.util.stream.Collectors;
+import java.util.Objects;
 
 /**
  * Enumeration of all the fields on a Claimant that may be updated by the claimant.
  */
+@Getter
 public enum UpdatableClaimantFields {
 
-    EXPECTED_DELIVERY_DATE("expectedDeliveryDate"),
-    ADDRESS("address") {
-
+    EXPECTED_DELIVERY_DATE("expectedDeliveryDate") {
         @Override
-        boolean valueIsDifferent(Claimant originalClaimant, Claimant newClaimant) throws InvocationTargetException, IllegalAccessException {
-            return hasAddressChanged(originalClaimant, newClaimant);
+        Object getValue(Claimant claimant) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFields.java
```diff
@@ -1,16 +1,107 @@
...
 /**
  * Enumeration of all the fields on a Claimant that may be updated by the claimant.
  */
-@AllArgsConstructor
-@Getter
 public enum UpdatableClaimantFields {
 
-    EXPECTED_DELIVERY_DATE("expectedDeliveryDate");
+    EXPECTED_DELIVERY_DATE("expectedDeliveryDate"),
+    ADDRESS("address") {
```
YetAnotherMatt: I added address as it needs special handling. We'd also add other fields such as name etc

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFields.java
```diff
@@ -1,16 +1,107 @@
...
+    private final Method setter;
+    private final Method getter;
+
+    /**
+     * Constructor that accepts the field name and uses it to obtain the getter and setter methods.
+     * @param fieldName the name of the field.
+     */
+    UpdatableClaimantFields(String fieldName) {
+        this.fieldName = fieldName;
+        Map<String, Method> methodMap = Arrays.stream(Claimant.class.getDeclaredMethods()).collect(Collectors.toMap(
```
YetAnotherMatt: I tried to make this map static, but it couldn't then be referred to from an enum constructor

mikeSmith08:
My thoughts on this are:
 - using reflection to build a map of all the getters and setters based on their name is quite clever, but seems overly complicated compared to the alternative which is to simply define them. I know this might become a maintenance nightmare, but is that any worse than complicated reflection code, plus you've already got that for the address anyway? Very much playing devil's advocate here, but I thought it would be good to say it anyway.
 - my main concern is that I don't like the main usage in the Production code in `NewClaimService`. I think it's a bit magical that the enum is doing the updating to the object as well as doing the comparison in one foul swoop. I think if we had something like:
   ```
if (isClaimantUpdated(originalClaimant, claimant)) {
    updateOriginalClaimant(originalClaimant, claimant);
}
```
I realise that this would need a lot of change from what you've written, but I worry that its too complicated for its own good when a simpler solution might suffice. I would love to pair through this with you and see what the simpler alternative would look like, but I realise that time is against me here so will go with whatever you decide.

YetAnotherMatt:
@dwybourn on Mike's suggestion I've removed the reflection and moved the control flow to NewClaimService. I think it probably is better for it. Have a look and let me know what you think, then I'll close this PR and put some tests in, and the other fields, when I get a chance.

PR #208, opened 2019-06-03
Feature/audit failed payment

mikeSmith08:
- Make sure that an EventFailedException is thrown for both makeFirstPayment and makePayment
 - Make sure that all available data that is available can be set on failed event (hence why the variables are defined outside of the try block and making PMD unhappy).

Also spotted a few things from my previous PR that could be improved: 
 - Make sure that we set the `cardAccountId` when `createNewCard` fails when saving the claim to the database.
 - Also spotted that we should be using the `createMakePaymentMessage` method in `DetermineEntitlementMessageProcessor` instead of calling the queue directly.
 - Improved some of the tests too, notably `NewCardServiceTest` and `EventFailedExceptionTest`.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/MakePaymentEvent.java
```diff
@@ -26,14 +23,14 @@ public MakePaymentEvent(UUID claimId, UUID paymentId, String reference, int paym
     private static Map<String, Object> constructMetaData(UUID claimId,
                                                          UUID paymentId,
                                                          String reference,
-                                                         int paymentAmountInPence,
-                                                         int entitlementAmountInPence) {
-        return Map.of(
```
dwybourn: Is this change to avoid NPEs?

mikeSmith08: Yep.

PR #211, opened 2019-06-03
HTBHF-1479 Make sure that a payment is saved for both first payments and normal payments should any failure in the event be detected.

mikeSmith08:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageTypeProcessor.java
```diff
@@ -22,4 +23,15 @@
...
     MessageType supportsMessageType();
 
+    /**
+     * Can be implemented by a MessageTypeProcessor to define custom functionality that needs to be performed
+     * when the processing of a message has failed. Default functionality as defined here is to do nothing.
+     *
+     * @param message      The message that has failed
+     * @param failureEvent The details of the failed event.
+     */
+    @SuppressWarnings("PMD.UncommentedEmptyMethodBody")
```
mikeSmith08: Its annoying I had to add this because its JavaDoc'd but PMD can't see it for some reason.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/MakePaymentMessageProcessor.java
```diff
@@ -40,4 +41,11 @@ public MessageType supportsMessageType() {
         return MAKE_PAYMENT;
     }
 
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public void processFailedMessage(Message message, FailureEvent failureEvent) {
```
mikeSmith08: This (and its test) are identical in both payment processors so I was wondering whether they should have some shared inheritance. Any thoughts? At least it should be in another PR if that's the case so I've left it for now.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentService.java
```diff
@@ -34,6 +40,35 @@
...
+                    .claim(paymentCycle.getClaim())
+                    .paymentAmountInPence(amountToPayInPence)
+                    .paymentCycle(paymentCycle)
+                    .paymentReference(paymentReference)
+                    .paymentStatus(PaymentStatus.FAILURE)
+                    .build();
+
+            paymentRepository.save(failedPayment);
+        } catch (Exception e) {
+            log.error("Unexpected exception caught saving a failed payment for paymentCycle: {}, cardAccountId: {}, failureEvent: {}, exception detail: {}",
```
mikeSmith08: I didn't see what else we could do should this attempt to save the failed payment also fail.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageTypeProcessor.java
YetAnotherMatt: We can remove the `@SuppressWarnings` and add a `// No action required` comment in the method body itself.

api/src/main/resources/application.yml
```diff
@@ -72,7 +72,7 @@ payment-cycle:
     weeks-after-due-date: 8
   schedule:
     # run at half-past the hour from 02:30 TO 05:30
-    cron-schedule: "0 30 2-5 * * *"
+    cron-schedule: "30 0 * * * *"
```
YetAnotherMatt: Is this 30 seconds past each hour?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentServiceTest.java
```diff
@@ -48,32 +54,48 @@
 import static uk.gov.dhsc.htbhf.logging.event.FailureEvent.FAILED_EVENT_KEY;
 import static uk.gov.dhsc.htbhf.logging.event.FailureEvent.FAILURE_DESCRIPTION_KEY;
 
-@ExtendWith(MockitoExtension.class)
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
```
YetAnotherMatt: Remind me why this needs to be a SpringBootTest? In general we need to reduce the number of SpringBootTests, as our build times are getting dangerously long.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageTypeProcessor.java
dwybourn: Removed suppress warnings.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentServiceTest.java
dwybourn: So we know that we're successfully creating entities in a valid state before persisting them.

api/src/main/resources/application.yml
dwybourn: Accidental check in, now reverted.

PR #212, opened 2019-06-03
HTBHF-1468 audit a failed claim event before the failed claim is saved

YetAnotherMatt:
I'm aware that this will conflict with another PR :(


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewClaimService.java
```diff
@@ -63,11 +65,27 @@ public ClaimResult createOrUpdateClaim(Claimant claimant) {
...
+                .claimId(claim.getId())
+                .claimStatus(claim.getClaimStatus())
+                .eligibilityStatus(claim.getEligibilityStatus())
+                .build();
+        FailureEvent failureEvent = FailureEvent.builder()
+                .failureDescription("Unable to create (or update) claim")
+                .failedEvent(newClaimEvent)
+                .exception(e)
+                .build();
+        eventAuditor.auditFailedEvent(failureEvent);
```
YetAnotherMatt: Corrected to audit a failed event instead of a new claim event, and ensured the audit happens before the save (as there's a good chance at this point that the failure to save the new claim was due to db errors).

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -62,11 +64,27 @@ public ClaimResult createOrUpdateClaim(Claimant claimant) {
...
         } catch (RuntimeException e) {
-            createAndSaveClaim(claimant, buildWithStatus(EligibilityStatus.ERROR));
+            handleFailedClaim(claimant, e);
             throw e;
         }
     }
 
+    private void handleFailedClaim(Claimant claimant, RuntimeException e) {
+        Claim claim = buildClaim(claimant, buildWithStatus(EligibilityStatus.ERROR));
+        NewClaimEvent newClaimEvent = NewClaimEvent.builder()
```
dwybourn: A constructor that takes in a Claim might look nicer here.

PR #221, opened 2019-06-10
HTBHF-1399 Added phone number to claimant in database.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claimant.java
```diff
@@ -45,4 +45,7 @@
     @NotNull
     @OneToOne(cascade = CascadeType.ALL)
     private Address address;
+
+    @Column(name = "phone_number")
```
YetAnotherMatt: Should we add validation here? We could add a regex pattern that enforces '+44' followed by 9 or 10 digits

dwybourn: I was planning on adding that in the next PR as we'll need NotNull on it as well. Potentially may use https://github.com/google/libphonenumber as recommended by govUK, but a simple regex may suffice.

mikeSmith08: We might want to be a bit more lenient on the phone number, there's a fair chance that some of the applicants might enter a non UK phone number.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/ClaimantTestDataFactory.java
```diff
@@ -17,6 +17,7 @@
     private static final String VALID_FIRST_NAME = "James";
     private static final String VALID_LAST_NAME = "Smith";
     private static final LocalDate VALID_DOB = LocalDate.parse("1987-12-30");
+    private static final String VALID_PHONE_NUMBER = "+447700900128";
```
mikeSmith08: I notice that above and here we are talking about phone numbers starting with +44, is this something we are asking the applicants to do, I'd expect them to simply start with 0, or are we converting it to have the international code? Seems a bit odd to me.

db/src/main/resources/db/migration/V1_028__add_phone_number_to_claimant.sql
```diff
@@ -0,0 +1 @@
+alter table claimant add column phone_number varchar(13);
```
mikeSmith08: Should this be `NOT NULL`, or is that happening in another PR?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claimant.java
gethyn1: In the kick off I think we agreed that it had to be a UK number

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/ClaimantTestDataFactory.java
dwybourn: In the kick off we decided we'd store the number in +44 format for consistency. We chose +44 over 0 because one of the card providers said they required numbers in that format.

db/src/main/resources/db/migration/V1_028__add_phone_number_to_claimant.sql
dwybourn: That'll be in the next PR as adding notnull now will cause the tests to fail. A TODO has been added to ClaimantTest linking to the next subtask.

PR #222, opened 2019-06-10
HTBHF-1533 upgraded gradle wrapper & spotbugs, adding spotbugs security plugin

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -123,6 +124,7 @@ private void sendNewCardMessage(Claim claim, EligibilityAndEntitlementDecision d
         messageQueueDAO.sendMessage(payload, CREATE_NEW_CARD);
     }
 
+    @SuppressFBWarnings("SECMD5") // ignore spotbugs warning about md5 as we're not using it for encryption
```
mikeSmith08: Can we get a bit more info on what was causing this and why?

dwybourn: It's for hashing the device fingerprint, so we can do comparisons against the hash rather than the whole thing.

dwybourn: Rather than a comment, this annotation takes in a justification value you can use instead.

PR #223, opened 2019-06-11
Feature/updatable claim fields

YetAnotherMatt:
We put this on hold pending clarification from Karen and review by Mike. Karen has confirmed that the list of fields that can be updated is correct (but will include phone number and email). 

See also: https://helptobuyhealthyfood.atlassian.net/browse/HTBHF-630?focusedCommentId=11861&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11861

@dwybourn you'll need to add a phone number enum value as part of the story you're working on.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantField.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantField.java
@@ -10,8 +10,41 @@ import java.util.Objects;
  * Enumeration of all the fields on a Claimant that may be updated by the claimant.
  */
 @Getter
-public enum UpdatableClaimantFields {
+public enum UpdatableClaimantField {
 
+    FIRST_NAME("firstName") {
+        @Override
+        Object getValue(Claimant claimant) {
+            return claimant.getFirstName();
+        }
+
+        @Override
+        public void updateOriginal(Claimant originalClaimant, Claimant newClaimant) {
+            originalClaimant.setFirstName(newClaimant.getFirstName());
+        }
+    },
+    LAST_NAME("lastName") {
+        @Override
+        Object getValue(Claimant claimant) {
+            return claimant.getLastName();
+        }
+
+        @Override
+        public void updateOriginal(Claimant originalClaimant, Claimant newClaimant) {
+            originalClaimant.setLastName(newClaimant.getLastName());
+        }
+    },
+    DATE_OF_BIRTH("dateOfBirth") {
+        @Override
+        Object getValue(Claimant claimant) {
+            return claimant.getDateOfBirth();
+        }
+
+        @Override
+        public void updateOriginal(Claimant originalClaimant, Claimant newClaimant) {
+            originalClaimant.setDateOfBirth(newClaimant.getDateOfBirth());
+        }
+    },
     EXPECTED_DELIVERY_DATE("expectedDeliveryDate") {
         @Override
         Object getValue(Claimant claimant) {
@@ -63,14 +96,15 @@ public enum UpdatableClaimantFields {
 
     private final String fieldName;
 
-    UpdatableClaimantFields(String fieldName) {
+    UpdatableClaimantField(String fieldName) {
         this.fieldName = fieldName;
     }
 
     /**
      * Checks whether the value of the relevant field differs between the two objects.
+     *
      * @param originalClaimant the original claimant.
-     * @param newClaimant the new claimant.
+     * @param newClaimant      the new claimant.
      * @return true if the new claimant has a different value to the original.
      */
     public boolean valueIsDifferent(Claimant originalClaimant, Claimant newClaimant) {
@@ -81,8 +115,9 @@ public enum UpdatableClaimantFields {
 
     /**
      * Updates the field value of the original to be the same as the new claimant.
+     *
      * @param originalClaimant the original claimant.
-     * @param newClaimant the new claimant.
+     * @param newClaimant      the new claimant.
      */
     public abstract void updateOriginal(Claimant originalClaimant, Claimant newClaimant);
 
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantField.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantField.java
@@ -56,6 +56,17 @@ public enum UpdatableClaimantField {
             originalClaimant.setExpectedDeliveryDate(newClaimant.getExpectedDeliveryDate());
         }
     },
+    PHONE_NUMBER("phoneNumber") {
+        @Override
+        Object getValue(Claimant claimant) {
+            return claimant.getPhoneNumber();
+        }
+
+        @Override
+        public void updateOriginal(Claimant originalClaimant, Claimant newClaimant) {
+            originalClaimant.setPhoneNumber(newClaimant.getPhoneNumber());
+        }
+    },
     ADDRESS("address") {
         @Override
         public boolean valueIsDifferent(Claimant originalClaimant, Claimant newClaimant) {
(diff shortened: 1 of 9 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantField.java
```diff
@@ -0,0 +1,125 @@
...
+import uk.gov.dhsc.htbhf.claimant.entity.Address;
+import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
+
+import java.util.Objects;
+
+/**
+ * Enumeration of all the fields on a Claimant that may be updated by the claimant.
+ */
+@Getter
+public enum UpdatableClaimantField {
```
mikeSmith08: Yes, this is much better and clearer than the first attempt using reflection, I like this! :-)

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -0,0 +1,124 @@
...
+        assertThat(UpdatableClaimantField.ADDRESS.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
+    }
+
+    @Test
+    void shouldReportFieldsDoNotHaveDifferentValues() {
+        LocalDate expectedDeliveryDate = LocalDate.now();
+        Claimant originalClaimant = aClaimantWithExpectedDeliveryDate(expectedDeliveryDate);
+        Claimant newClaimant = aClaimantWithExpectedDeliveryDate(expectedDeliveryDate);
+
+        assertThat(UpdatableClaimantField.FIRST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
```
mikeSmith08: Might it be worthwhile in these tests (especially this one where all the fields in the two objects are the same) to initially put an assertion to make sure the two objects aren't equal? Not essential but I thought it would just make it clear that the two objects aren't the same in all the tests.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -0,0 +1,124 @@
...
+import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
+
+import java.time.LocalDate;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.AddressTestDataFactory.aValidAddress;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.AddressTestDataFactory.aValidAddressBuilder;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.ClaimantTestDataFactory.*;
+
+class UpdatableClaimantFieldTest {
```
mikeSmith08: One thing you could do for bonus credit (obviously optional) is to create a custom assertion class like `ConstraintViolationAssert` and have your assertions read something like: `assertThat(originalClaimant).hasDifferentFirstName();` and `assertThat(originalClaimant).hasNoDifferencesTo(newClaimant)` etc. The tests would read really nicely then, but at current would only help this unit test, so up to you whether you feel like doing it.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
YetAnotherMatt: Last time we did that David raised (valid) questions about having an assert in the given step

mikeSmith08: Which I disagreed with last time too, I think its valid to assert that your input data is in the correct state before you start the test. Maybe you could also validly say that it should be in the then step as it is an assertion.

YetAnotherMatt: I've added the assert

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -18,36 +18,32 @@ void shouldReportFieldsHaveDifferentValues() {
         Claimant originalClaimant = Claimant.builder().build();
         Claimant newClaimant = aClaimantWithExpectedDeliveryDate(LocalDate.now());
 
-        assertThat(UpdatableClaimantField.FIRST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
-        assertThat(UpdatableClaimantField.LAST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
-        assertThat(UpdatableClaimantField.DATE_OF_BIRTH.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
-        assertThat(UpdatableClaimantField.EXPECTED_DELIVERY_DATE.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
-        assertThat(UpdatableClaimantField.ADDRESS.valueIsDifferent(originalClaimant, newClaimant)).isTrue();
+        for (UpdatableClaimantField field : UpdatableClaimantField.values()) {
```
mikeSmith08: Could this be an `@EnumSource`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -18,36 +18,32 @@ void shouldReportFieldsHaveDifferentValues() {
...
         Claimant originalClaimant = aClaimantWithExpectedDeliveryDate(expectedDeliveryDate);
         Claimant newClaimant = aClaimantWithExpectedDeliveryDate(expectedDeliveryDate);
+        assertThat(newClaimant).isNotEqualTo(originalClaimant);
 
-        assertThat(UpdatableClaimantField.FIRST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.LAST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.DATE_OF_BIRTH.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.EXPECTED_DELIVERY_DATE.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.ADDRESS.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
+        for (UpdatableClaimantField field : UpdatableClaimantField.values()) {
```
mikeSmith08: Same here, could this be an @EnumSource?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -18,36 +18,32 @@ void shouldReportFieldsHaveDifferentValues() {
...
         Claimant originalClaimant = Claimant.builder().build();
         Claimant newClaimant = Claimant.builder().build();
+        assertThat(newClaimant).isNotEqualTo(originalClaimant);
 
-        assertThat(UpdatableClaimantField.FIRST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.LAST_NAME.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.DATE_OF_BIRTH.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.EXPECTED_DELIVERY_DATE.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
-        assertThat(UpdatableClaimantField.ADDRESS.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
+        for (UpdatableClaimantField field : UpdatableClaimantField.values()) {
```
mikeSmith08: Ditto

PR #227, opened 2019-06-17
HTBHF-1600 Calculate interim pregnancy payment vouchers.

dwybourn:
Calculates the number of pregnancy vouchers a claimant is entitled to when they have updated their claim to include an expected delivery date. See https://helptobuyhealthyfood.atlassian.net/browse/HTBHF-1193 for full acceptance criteria.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param expectedDueDate       the expected due date of the claimant
+     * @param paymentCycleStartDate the start date for the current payment cycle
+     * @param claimUpdatedDate      the date that the claim was updated with the new expected due date
+     * @return the number of additional pregnancy vouchers the claimant is entitled to
+     */
+    public int getAdditionalPregnancyVouchers(LocalDate expectedDueDate,
+                                              LocalDate paymentCycleStartDate,
+                                              LocalDate claimUpdatedDate) {
+        if (claimUpdatedDate.isBefore(paymentCycleStartDate)) {
```
YetAnotherMatt: Could we just handle this and the check below as
```
        if (!claimUpdatedDate.isAfter(paymentCycleStartDate)) {
            // if the update happened on the same day as the payment cycle start date,
            // then no need for additional vouchers as the payment cycle will now include pregnancy vouchers.
            return 0;
        }
```
?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
```diff
@@ -0,0 +1,83 @@
...
+                .map(entitlementDate -> pregnancyEntitlementCalculator.isEntitledToVoucher(expectedDueDate, entitlementDate))
+                .mapToInt(isEntitled -> isEntitled ? vouchersPerPregnancy : 0)
+                .sum();
+
+        return Math.toIntExact(pregnancyVouchers);
+    }
+
+    private List<LocalDate> getEntitlementDates(LocalDate paymentCycleStartDate, LocalDate claimUpdatedDate) {
+        LocalDate rollingEntitlementDate = getFirstEntitlementDateOnOrAfterClaimUpdatedDate(paymentCycleStartDate, claimUpdatedDate);
+        LocalDate cycleEndDate = paymentCycleStartDate.plusDays(entitlementCalculationDurationInDays * numberOfCalculationPeriods);
```
YetAnotherMatt: Can we pass in both cycleStartDate and cycleEndDate rather than calculating the end date? I know they _should_ be the same, but I'd rather rely on the cycleEndDate as it is specified in the database. Just in case at some point we find we need to create a short or long cycle (e.g. if we're transitioning users automatically and need to spread them across the existing 28 day cycle)

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public MessageStatus processMessage(Message message) {
+        AdditionalPregnancyPaymentMessageContext context = messageContextLoader.loadAdditionalPregnancyPaymentMessageContext(message);
+        Optional<PaymentCycle> paymentCycle = context.getPaymentCycle();
+        if (paymentCycle.isEmpty() || hasPregnancyVouchers(paymentCycle)) {
+            return COMPLETED;
+        }
+
```
YetAnotherMatt: We need a TODO here at least, to signify we need to calculate entitlement and make payment

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+        assertThat(result).isEqualTo(expectedNumberOfVouchers);
+        ArgumentCaptor<LocalDate> argumentCaptor = ArgumentCaptor.forClass(LocalDate.class);
+        verify(pregnancyEntitlementCalculator, times(expectedNumberOfVouchers)).isEntitledToVoucher(eq(expectedDueDate), argumentCaptor.capture());
+        List<LocalDate> entitlementDates = argumentCaptor.getAllValues();
+        assertThat(entitlementDates).hasSize(expectedNumberOfVouchers);
+    }
+
+    @Test
+    void shouldThrowExceptionWhenClaimUpdatedDateIsBeforePaymentCycleStartDate() {
```
YetAnotherMatt: I'm not sure we should be throwing an exception in this case. Isn't it sufficient to simply not make a payment (i.e. return 0 entitlement)? Up for debate, but what happens if this exception is thrown? We'll never process this message.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: Replaced both checks with the one above.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: Passed in PaymentCycle instead which includes both start and end dates.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
dwybourn: Added TODO

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
dwybourn: Now just returning 0.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
```diff
@@ -0,0 +1,77 @@
...
+            return 0;
+        }
+
+        List<LocalDate> entitlementDates = getEntitlementDates(paymentCycle, claimUpdatedDate);
+        long pregnancyVouchers = entitlementDates.stream()
+                .map(entitlementDate -> pregnancyEntitlementCalculator.isEntitledToVoucher(expectedDueDate, entitlementDate))
+                .mapToInt(isEntitled -> isEntitled ? vouchersPerPregnancy : 0)
+                .sum();
+
+        return Math.toIntExact(pregnancyVouchers);
```
YetAnotherMatt: Is there any need to make `pregnancyVouchers` a long, then use `Math.toIntExact`? I think `mapToInt` returns an int, so could we make `pregnancyVouchers` an int too?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+     * @param message The message to process.
+     * @return The message status on completion
+     */
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public MessageStatus processMessage(Message message) {
+        AdditionalPregnancyPaymentMessageContext context = messageContextLoader.loadAdditionalPregnancyPaymentMessageContext(message);
+        Optional<PaymentCycle> paymentCycle = context.getPaymentCycle();
+        if (paymentCycle.isEmpty() || hasPregnancyVouchers(paymentCycle)) {
+            return COMPLETED;
```
YetAnotherMatt: might want a log.debug statement here

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: That's left over from when I implemented it a different way which returned a long. Removed now.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
dwybourn: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
```diff
@@ -0,0 +1,75 @@
...
+            // if the update happened on the same day as the payment cycle start date,
+            // then no need for additional vouchers as the payment cycle will now include pregnancy vouchers.
+            return 0;
+        }
+
+        List<LocalDate> entitlementDates = getEntitlementDates(paymentCycle, claimUpdatedDate);
+        return entitlementDates.stream()
+                .map(entitlementDate -> pregnancyEntitlementCalculator.isEntitledToVoucher(expectedDueDate, entitlementDate))
+                .mapToInt(isEntitled -> isEntitled ? vouchersPerPregnancy : 0)
+                .sum();
```
mikeSmith08: Just as an alternative, doesn't this read clearer?
`
List<LocalDate> entitlementDates = getEntitlementDates(paymentCycle, claimUpdatedDate);
        int totalVouchers = 0;
        for (LocalDateTime entitlementDate : entitlementDates) {
            if(pregnancyEntitlementCalculator.isEntitledToVoucher(expectedDueDate, entitlementDate)) {
                totalVouchers += vouchersPerPregnancy;
            }
        }
        return totalVouchers;
`
This seems simpler to me rather than using stream, map, mapToInt and sum for a simple calculation. Thoughts?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param message The message to process.
+     * @return The message status on completion
+     */
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public MessageStatus processMessage(Message message) {
+        AdditionalPregnancyPaymentMessageContext context = messageContextLoader.loadAdditionalPregnancyPaymentMessageContext(message);
+        Optional<PaymentCycle> paymentCycle = context.getPaymentCycle();
+        if (paymentCycle.isEmpty() || hasPregnancyVouchers(paymentCycle)) {
+            log.debug("Making no additional payments as the is no payment cycle or payment cycle already contains pregnancy vouchers.");
```
mikeSmith08: Should be `as there is no payment cycle`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
YetAnotherMatt: I almost made the same comment, then decided I was just being old-fashioned. I agree though, I find the for loop clearer.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
dwybourn: Fixed.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: I prefer the stream but can change if others prefer.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+            return COMPLETED;
+        }
+
+        // HTBHF-1193 TODO calculate entitlement and make payment
```
mikeSmith08: I was initially confused by this as it seemed to do nothing, so I presume that you're going to use the calculator in the next PR? If I was going to do that I'd have probably added the calculator in that PR rather than this one, I was expecting it to be used.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @ParameterizedTest
+    @CsvSource({
+            "-5, 0", // no vouchers if the claim updated date is before the payment cycle start date
+            "0, 0",  // if on the same day as the payment cycle then no need for additional vouchers as the payment cycle will now include pregnancy vouchers.
+            "1, 3",  // week one, get vouchers for week two, three and four
+            "8, 2",  // week two, get vouchers for week three and four
+            "15, 1", // week three, get vouchers for week four
+            "22, 0"  // week four, get no vouchers
+    })
+    void shouldCalculateNumberOfAdditionalPregnancyVouchers(int daysInCycle, int expectedNumberOfVouchers) {
```
mikeSmith08: The variable `daysInCycle` doesn't make sense, especially as the first value is -5, you can't have a negative cycle length.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+        lenient().when(pregnancyEntitlementCalculator.isEntitledToVoucher(any(), any())).thenReturn(true);
+        PaymentCycle paymentCycle = aPaymentCycleWithStartAndEndDate(paymentCycleStartDate, paymentCycleEndDate);
+
+        int result = calculator.getAdditionalPregnancyVouchers(expectedDueDate, paymentCycle, claimUpdatedDate);
+
+        assertThat(result).isEqualTo(expectedNumberOfVouchers);
+        ArgumentCaptor<LocalDate> argumentCaptor = ArgumentCaptor.forClass(LocalDate.class);
+        verify(pregnancyEntitlementCalculator, times(expectedNumberOfVouchers)).isEntitledToVoucher(eq(expectedDueDate), argumentCaptor.capture());
+        List<LocalDate> entitlementDates = argumentCaptor.getAllValues();
+        assertThat(entitlementDates).hasSize(expectedNumberOfVouchers);
```
mikeSmith08: This should be `assertThat(entitlementDates).hasSize(expectedNumberOfVouchers * VOUCHERS_PER_PREGNANCY)`, it just happens that VOUCHERS_PER_PREGNANCY is 1, but we shouldn't make the assumption that we know the value.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoaderTest.java
```diff
@@ -355,4 +355,28 @@ void shouldSuccessfullyLoadNewCardContext() {
...
 
+    @Test
+    void shouldSuccessfullyLoadAdditionalPregnancyPaymentContext() {
+        //Given
+        Claim claim = aValidClaim();
+        UUID claimId = claim.getId();
+        PaymentCycle latestPaymentCycle = aPaymentCycleWithClaim(claim);
+        given(claimRepository.findById(any())).willReturn(Optional.of(claim));
+        given(paymentCycleRepository.findCurrentCycleForClaim(any())).willReturn(Optional.of(latestPaymentCycle));
+        var payload = AdditionalPregnancyPaymentMessagePayload.builder().claimId(claimId).build();
```
mikeSmith08: Shouldn't be using `var` here.

mikeSmith08: Does this warrant a Test Data Factory? We might need one when you build on this in the next PR(s).

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
YetAnotherMatt: it's a fair point. The calculator is definitely included in the next PR.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param expectedDueDate  the expected due date of the claimant
+     * @param paymentCycle     the current payment cycle
+     * @param claimUpdatedDate the date that the claim was updated with the new expected due date
+     * @return the number of additional pregnancy vouchers the claimant is entitled to
+     */
+    public int getAdditionalPregnancyVouchers(LocalDate expectedDueDate,
+                                              PaymentCycle paymentCycle,
+                                              LocalDate claimUpdatedDate) {
+        if (!claimUpdatedDate.isAfter(paymentCycle.getCycleStartDate())) {
+            // if the update happened on the same day as the payment cycle start date,
```
mikeSmith08: Is there the possibility that if the request is on the same date as the cycle start date that it might happen after we've already made the calculation and payment, so might have missed the payment and not get paid? Could be a timing issue here.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoaderTest.java
dwybourn: I don't seem the harm here when we can see that we're calling build on the same line. It's particular nice when you have very long class names.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
dwybourn: That's not right, the entitlement dates are the dates to calculate entitlement for, not the total number of vouchers.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessorTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Mock
+    private MessageContextLoader messageContextLoader;
+    @Mock
+    private PaymentService paymentService;
+
+    @InjectMocks
+    private AdditionalPregnancyPaymentMessageProcessor messageProcessor;
+
+    @Test
+    void shouldProcessMessageAndMakePaymentWhenCurrentPaymentCycleHasNoPregnancyVouchers() {}
```
mikeSmith08: This should either have an implementation or be `@Disabled`, we shouldn't have tests that are Green that do nothing.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
dwybourn: It's the day after the cycle was created, the comments should explain but I'll rename the variable as well.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessor.java
dwybourn: Adding to next PR which actually does the payment. Otherwise it's calling a calculator and doing nothing with the result

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: @YetAnotherMatt Did we discuss this scenario?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/AdditionalPregnancyPaymentMessageProcessorTest.java
dwybourn: Added TODO

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
dwybourn: That's not right, the entitlement dates are the dates to calculate entitlement for, not the total number of vouchers.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
YetAnotherMatt: We discussed what would happen if there was no current paymentCycle. I think Mike is on to something here.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculatorTest.java
dwybourn: Updated to  
`assertThat(entitlementDates).hasSize(expectedNumberOfVouchers / VOUCHERS_PER_PREGNANCY)`

Should be '/' instead of '*'

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/AdditionalPregnancyVoucherCalculator.java
dwybourn: Updated with additional check to make sure the payment cycle has also not been processed already.

PR #229, opened 2019-06-19
HTBHF-1610 Added email address to the claimant.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
```diff
@@ -64,4 +64,16 @@
     @JsonProperty("phoneNumber")
     @ApiModelProperty(notes = "The claimant's UK phone number. Must be in +44 format, e.g. +447123456789")
     private String phoneNumber;
+
+    // See https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression for explanation of email regex.
+    @NotNull
+    @Pattern(regexp = "(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]"
```
YetAnotherMatt: suggest we have a single constant for this regex

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
```diff
@@ -223,4 +223,51 @@ void shouldValidateClaimantWithPhoneNumberInCorrectFormat(String phoneNumber) {
...
+            "email@domain.com",
+            "firstname.lastname@domain.com",
+            "email@subdomain.domain.com",
+            "firstname+lastname@domain.com",
+            "1234567890@domain.com",
+            "email@domain-one.com",
+            "_______@domain.com",
+            "email@domain.name",
+            "email@domain.co.jp",
+            "firstname-lastname@domain.com"
```
YetAnotherMatt: can we validate non-ascii characters?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
dwybourn: Done.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/Constants.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.dhsc.htbhf.claimant.model;
+
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+public final class Constants {
+
+    // Using simplified regex due to limitations of java unicode in regular expressions.
```
dwybourn: The regex is much simpler than the version used in the frontend due to difference in regex behaviour between JS and Java. The key thing is that all emails that are valid in the UI are also valid here.

PR #232, opened 2019-06-26
HTBHF-1588 Add date of birth of children to the Claimant table and as…

mikeSmith08:
… an optional request parameter for a new claim.

Only validation on the request is that if there are any dates provided, they all need to be in the past.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
@@ -8,8 +8,10 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import uk.gov.dhsc.htbhf.claimant.model.constraint.DateWithinRelativeRange;
+import uk.gov.dhsc.htbhf.claimant.model.constraint.ListOfDatesInPast;
 
 import java.time.LocalDate;
+import java.util.List;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Past;
@@ -74,5 +76,10 @@ public class ClaimantDTO {
     @JsonProperty("emailAddress")
     @ApiModelProperty(notes = "The claimant's email address. e.g. test@email.com")
     private String emailAddress;
+
+    @JsonProperty("childrenDob")
+    @ApiModelProperty(notes = "The dates of birth of the claimant's children (if they have any)")
+    @ListOfDatesInPast(message = "dates of birth of children should be all in the past")
+    private List<LocalDate> childrenDob;
 
 }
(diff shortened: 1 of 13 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
```diff
@@ -75,4 +76,8 @@
     @ApiModelProperty(notes = "The claimant's email address. e.g. test@email.com")
     private String emailAddress;
 
+    @JsonProperty("childrenDob")
+    @ApiModelProperty(notes = "The dates of birth of the claimant's children (if they have any)")
+    private List<LocalDate> childrenDob;
```
dwybourn: Should have ListOfDatesInPast annotation.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/constraint/ListOfDatesInPastValidator.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public boolean isValid(List<LocalDate> dates, ConstraintValidatorContext context) {
+        if (CollectionUtils.isEmpty(dates)) {
+            return true;
+        }
+        return allDatesInPast(dates);
+    }
+
+    private boolean allDatesInPast(List<LocalDate> dates) {
+        LocalDate now = LocalDate.now();
```
dwybourn: For loop and below can be replaced with a one liner `return dates.stream().noneMatch(now::isBefore);`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
```diff
@@ -298,4 +299,25 @@ void shouldValidateClaimantWithValidEmailAddress(String emailAddress) {
         //Then
         assertThat(violations).hasNoViolations();
     }
+
```
dwybourn: Can we have an empty list test as well.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimantDTO.java
mikeSmith08: Yep, good spot. This does show that we don't have unit tests for the ClaimDTO, I'll add in another PR.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/constraint/ListOfDatesInPastValidator.java
mikeSmith08: I like the simplicity of my version, so I'm going to keep it, this one makes me think too much for something that is very simple.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimantTest.java
mikeSmith08: The first test does that by using `aValidClaimant()` which has no children's date of births in it.

PR #234, opened 2019-06-27
Feature/add missing validation tests

mikeSmith08:
Add missing unit tests for the DTO object validation, plus remove those tests that would be duplicates from the integration tests.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/constraint/DateWithinRelativeRange.java
```diff
@@ -56,34 +53,3 @@
      */
     int monthsInFuture();
 }
-
-class DateWithinRelativeRangeValidator implements ConstraintValidator<DateWithinRelativeRange, LocalDate> {
```
YetAnotherMatt: I quite liked having the constraint and the validator in the same file! Not enough to argue about it though.

mikeSmith08: I had to do this because there is no way for the unit test to construct the validator otherwise, the only way this was working in tests previously was because you were using Mockito which can access the package private class.

PR #242, opened 2019-07-22
HTBHF-1882 Rename SEND_FIRST_EMAIL MessageType tpo SEND_EMAIL in preparation for email message processor.

mikeSmith08:
We don't need message types for each email message as we can easily do this with a single processor for all emails, so renaming to message type accordingly.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
```diff
@@ -4,7 +4,7 @@
     CREATE_NEW_CARD,
     MAKE_FIRST_PAYMENT,
     MAKE_PAYMENT,
-    SEND_FIRST_EMAIL,
+    SEND_EMAIL,
```
YetAnotherMatt: This will only work as long as the message is created with the template ID and all required variable values. This might mean that the method that invokes the sending of an email gets a bit too large. I guess we can track that as it happens - we can always reinstate 'send_first_email' and have it amass the required variables then trigger 'send_email' if we need to.

mikeSmith08: I'm going to create a configuration class that stores the template ids and have an enum that the email messages reference instead, when the message context is inflated from the payload we can then lookup the templateId from the enum value. That was my thoughts anyway.

YetAnotherMatt: Sounds good. It was more knowing which variables to obtain that I was concerned about.

PR #243, opened 2019-07-22
HTBHF-1882 Create message context and payload for email messages, plus loading of context from payload.

mikeSmith08:
- Create payload and context objects for email message
 - Load `EmailMessageContext` from `EmailMessagePayload`
 - Added basic implementation of how the email message type -> templateId might be configured and stored to get the MessageContextLoader working.

I implemented the first (and simplest thing) that came to mind on how to configure the template ids by message type so we can reuse this for all future messages, so any alternate thoughts welcome. The idea is that we'd simple need to add a new value to the `EmailType` enum, a new config value in `application.yml` and then add to `EmailMessageConfiguration`. The last bit is a bit clunky, but I'm sure that will be iterated upon in the next PR or two.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailMessageConfiguration.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailMessageConfiguration.java
@@ -1,0 +1,27 @@
+package uk.gov.dhsc.htbhf.claimant;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import uk.gov.dhsc.htbhf.claimant.message.payload.EmailType;
+
+import java.util.Map;
+
+@EnableConfigurationProperties
+@Configuration
+@ConfigurationProperties(prefix = "notify")
+@Getter
+@Setter
+public class EmailMessageConfiguration {
+
+    private Map<EmailType, String> templateIds;
+
+    @Bean
+    public EmailTemplateConfig emailTemplateConfig() {
+        return new EmailTemplateConfig(templateIds);
+    }
+
+}
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailTemplateConfig.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailTemplateConfig.java
@@ -1,0 +1,31 @@
+package uk.gov.dhsc.htbhf.claimant;
+
+import uk.gov.dhsc.htbhf.claimant.message.payload.EmailType;
+
+import java.util.Map;
+
+/**
+ * Contains all the mappings between email types and template ids. This is built on startup
+ * from configuration only.
+ */
+public class EmailTemplateConfig {
+
+    private final Map<EmailType, String> templateIds;
+
+    public EmailTemplateConfig(Map<EmailType, String> templateIds) {
+        this.templateIds = templateIds;
+    }
+
+    /**
+     * Get the templateId for the given email type.
+     *
+     * @param emailType The type of email to get the template id for
+     * @return The template id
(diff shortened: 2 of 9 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailTemplateConfig.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Map;
+
+/**
+ * Contains all the mappings between email types and template ids. This is built on startup
+ * from configuration only.
+ */
+public class EmailTemplateConfig {
+
+    private final Map<EmailType, String> templateIds = new HashMap<>();
+
```
YetAnotherMatt: We could consider initialising `EmailTemplateConfig` with a map obtained from yml:
```
public EmailTemplateConfig(@Value("${notify.template-ids") Map<EmailType, String> templateIds) {
    this.templateIds = templateIds;
}
```
And in the yml:
```
notify:
  template-ids:
    NEW_CARD: bbbd8805-b020-41c9-b43f-c0e62318a6d5
```
I've not tried this, but I have a feeling it might work, fingers crossed.

mikeSmith08: That's what I was hoping we might be able to do, didn't know if this worked or not. I'll give it a try.

mikeSmith08: Ok,  got this working by making `EmailMessageConfiguration` a `@ConfigurationProperties`, we can simply add template ids and email types easily then.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailMessageConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.context.annotation.Configuration;
+import uk.gov.dhsc.htbhf.claimant.message.payload.EmailType;
+
+import java.util.Map;
+
+@EnableConfigurationProperties
+@Configuration
+@ConfigurationProperties(prefix = "notify")
+@Getter
+@Setter
```
mikeSmith08: In case you're wondering why this is here, the `Map` doesn't get populated without a getter and setter.

YetAnotherMatt: Minor, but `@Getter` and `@Setter` could maybe go on the Map itself

api/src/test/java/uk/gov/dhsc/htbhf/claimant/EmailTemplateConfigTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    void shouldFailToFindTemplateIdIfNotStored() {
+        //Given no templateIds stored
+        //When
+        IllegalArgumentException caught = catchThrowableOfType(
+                () -> emailTemplateConfig.getTemplateIdForEmail(EmailType.PAYMENT),
+                IllegalArgumentException.class);
+        //Then
+        assertThat(caught).hasMessage("No template id defined for email type: PAYMENT");
```
YetAnotherMatt: This seems a bit temporary (we'll eventually have templates for all email types). Perhaps we should instead have a parameterised test with an EnumSource that confirms that all email types have a template id? This might need to be a spring boot test, so we can obtain the correct config.

mikeSmith08: In the context of the test, this is fine as the Map of templateIds is created in the test and controlled in the test. Seems overkill to make this a Spring Boot test when its simply testing that an Exception is thrown if we request a templateId for an EmailType that isn't configured.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/EmailMessageConfiguration.java
mikeSmith08: Ok, will do this on the next PR to save some build time.

PR #244, opened 2019-07-24
HTBHF-1882 Create EmailMessageProcessor which isn't used yet.

mikeSmith08:
This will be used in a subsequent PR but is plumbed into the context and available to `MessageProcessor`.

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/EmailMessageProcessorTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/EmailMessageProcessorTest.java
@@ -1,0 +1,68 @@
+package uk.gov.dhsc.htbhf.claimant.message.processor;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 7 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/EmailMessageProcessorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    void shouldSendMessage() throws NotificationClientException {
+        //Given
+        Claim claim = aValidClaim();
+        String templateId = "12334546";
+        EmailMessageContext context = EmailMessageContext.builder()
+                .claim(claim)
+                .templateId(templateId)
+                .emailPersonalisation(emptyMap())
```
YetAnotherMatt: Could we use a map with a key/value pair? It would be nice to assert that the correct map is passed to notify.

mikeSmith08: Fair point, have added a representative Map of data via a Test Data Factory (we'll almost certainly be using it in the next PR anyway).

PR #246, opened 2019-07-25
HTBHF-1883 Correct data setup in PaymentCycleVoucherEntitlementTestDataFactory…

mikeSmith08:
… so that each PaymentCycleVoucherEntitlement that is created is actually representative of a real payment cycle, in that we'll have 4 vouch entitlements each a week apart.

The next PR will then focus on the PaymentCycleTestDataFactory and make sure that the vouchers and values created are correct according to the vouchers that are contained within.


mikeSmith08:
I realise that there's plenty more to do on this TestDataFactory and its not quite right yet, so going to close this.

PR #256, opened 2019-08-05
HTBHF-1915 Added county field to address. Defaults to UK.

dwybourn:
Added county to address table. Defaults to UK as is required by AllPay but actual county doesn't matter.

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
@@ -38,6 +38,16 @@ public class AddressTest extends AbstractValidationTest {
     }
 
     @Test
+    void shouldValidateAddressSuccessfullyWithNoCounty() {
+        //Given
+        Address address = anAddressWithCounty(null);
+        //When
+        Set<ConstraintViolation<Address>> violations = validator.validate(address);
+        //Then
+        assertThat(violations).hasNoViolations();
+    }
+
+    @Test
     void shouldFailToValidateAddressWithNoAddressLine1() {
         //Given
         Address address = anAddressWithAddressLine1(null);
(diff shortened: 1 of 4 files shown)


db/src/main/resources/db/migration/V1_036__add_county_to_address.sql
```diff
@@ -0,0 +1 @@
+alter table address add column county varchar(500) default 'UK';
```
YetAnotherMatt: A default won't give existing rows a value. I wonder if it might be better to define a default when we send values to Allpay (and not in the table), but I guess that would make the sql extract quite awkward.

dwybourn: Default does give existing rows a value, just tested it locally. 
At the moment will set England as a value for County for AllPay when generating the csv. Surely it would make sense to use the value given if the user provide it.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
```diff
@@ -87,6 +87,16 @@ void shouldFailToValidateAddressWhenTownOrCityIsTooLong() {
         assertThat(violations).hasSingleConstraintViolation("size must be between 1 and 500", "townOrCity");
     }
 
+    @Test
+    void shouldFailToValidateAddressWhenCountyIsTooLong() {
```
mikeSmith08: Can we have a test that says there are no validations when the county is missing, or is that covered by the default success test here? Either way we should have a test with and without the county.

db/src/main/resources/db/migration/V1_036__add_county_to_address.sql
mikeSmith08: If we're using England at the moment, why are we defaulting to UK here?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/AddressTest.java
dwybourn: Done.

db/src/main/resources/db/migration/V1_036__add_county_to_address.sql
dwybourn: We decided to use UK as that's a better long term solution. The current allpay instructions are just for private beta, which is only users from England.

YetAnotherMatt: My thinking was that we'd default to 'UK' if there was no value here, rather than store a default of 'UK' in the table, but I think that would cause problems with our interim sql extract.

PR #258, opened 2019-08-20
HTBHF-2120 Using circleci

dwybourn:


.circleci/config.yml
```diff
@@ -0,0 +1,102 @@
...
+      - save_cache:
+          paths:
+            - ~/.gradle/wrapper
+          key: v1-gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+      - save_cache:
+          paths:
+            - ~/.gradle/caches
+          key: v1-gradle-cache-{{ checksum "build.gradle" }}
+      - run:
+          name: deploy
```
YetAnotherMatt: This should only happen on a deploy to master

PR #263, opened 2019-08-27
HTBHF-2171 Increased the timeout for the initial healthcheck from 1 to 3 minutes

YetAnotherMatt:
Increases the timeout for the first healthcheck (after startup) - the normal healthcheck is unchanged.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -5,6 +5,7 @@ applications:
   - java_buildpack
   health-check-type: http
   health-check-http-endpoint: /actuator/health
+  health-check-timeout: 180
   routes:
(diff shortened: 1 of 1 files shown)


manifest.yml
```diff
@@ -5,6 +5,7 @@ applications:
   - java_buildpack
   health-check-type: http
   health-check-http-endpoint: /actuator/health
+  health-check-timeout: 180
```
dwybourn: Will this be enough time on staging where we will be updating about half a million records?

YetAnotherMatt: This is the maximum we can configure without asking the PaaS team to extend it. The update has already happened as part of the failed build, so the next build won't take much time at all.

PR #267, opened 2019-09-09
NO-JIRA ensured postcode is alwasy uppercase. Moved some log to messages before calling db.

YetAnotherMatt:
The two things are unrelated. Uppercase postcode as discussed with Karen earlier today. Log messages because this helps track down issues when db access is slow.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/converter/AddressDTOToAddressConverterTest.java
```diff
@@ -28,6 +30,18 @@ void shouldConvertAddressDTOToEquivalentAddressObject() {
...
+    @Test
+    void shouldConvertPostcodeToUppercase() {
+        // Given
+        AddressDTO addressDTO = anAddressDTOWithPostcode("aa11aa");
+
+        // When
+        Address result = converter.convert(addressDTO);
+
+        // Then
+        assertThat(result.getPostcode()).isEqualTo(addressDTO.getPostcode().toUpperCase());
```
mikeSmith08: I'd prefer this to specifically say "AA11AA" as the expected value rather than take it from the DTO.

PR #269, opened 2019-09-16
HTBHF-2241 Added a PaymentNotification email message payload to MessagePayloadFactory

YetAnotherMatt:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactoryTest.java
```diff
@@ -73,4 +73,31 @@ void shouldBuildSendNewCardSuccessEmailPayloadWithOnlyPregnancyPayment() {
...
+
+        LocalDate startDate = LocalDate.now();
+        LocalDate endDate = startDate.plusDays(28);
+        PaymentCycle paymentCycle = aPaymentCycleWithStartAndEndDate(startDate, endDate);
+
+        EmailMessagePayload payload = MessagePayloadFactory.buildPaymentNotificationEmailPayload(paymentCycle);
+
+        assertThat(payload.getClaimId()).isEqualTo(paymentCycle.getClaim().getId());
+        assertThat(payload.getEmailType()).isEqualTo(EmailType.PAYMENT);
+        assertEmailPayloadCorrectForClaimantWithAllVouchers(payload.getEmailPersonalisation(), endDate);
```
mikeSmith08: This method doesn't check that the field `payment_amount` is present in the Map so that should be updated too.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactory.java
```diff
@@ -49,23 +49,39 @@ public static MakePaymentMessagePayload buildMakePaymentMessagePayload(PaymentCy
...
+
+    private static Map<String, Object> createPaymentEmailPersonalisationMap(PaymentCycle paymentCycle) {
         Claimant claimant = paymentCycle.getClaim().getClaimant();
         PaymentCycleVoucherEntitlement voucherEntitlement = paymentCycle.getVoucherEntitlement();
         Map<String, Object> emailPersonalisation = new HashMap<>();
         emailPersonalisation.put(EmailTemplateKey.FIRST_NAME.getTemplateKeyName(), claimant.getFirstName());
         emailPersonalisation.put(EmailTemplateKey.LAST_NAME.getTemplateKeyName(), claimant.getLastName());
-        emailPersonalisation.put(EmailTemplateKey.FIRST_PAYMENT_AMOUNT.getTemplateKeyName(),
-                convertPenceToPounds(paymentCycle.getTotalEntitlementAmountInPence()));
+        // TODO: MGS: HTBHF-2282 remove FIRST_PAYMENT_AMOUNT once the instant success template has been updated to use PAYMENT_AMOUNT
```
mikeSmith08: I thought the templates were ready to go? I don't like working on these changes when the template isn't 100% sure, I'd have kicked this back to Sam/Laura before starting if that was the case.

YetAnotherMatt: I was thinking we should change the instant success template (already in use) so that it uses 'payment_amount' instead of 'first_payment_amount' to be consistent with other templates. One alternative is to specify both, as we're doing here. Another is to have different methods populating the message payload for different notify templates.

PR #273, opened 2019-09-18
HTBHF-2288 Add method to MessagePayloadFactory to build payload for…

mikeSmith08:
…email when a claimant's child turns 4.

Not currently plugged in anywhere, that will be a subsequent PR. Will concentrate on the calculations around children's dates of birth in the next PR.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactory.java
@@ -17,6 +17,9 @@ import java.util.Map;
 
 import static uk.gov.dhsc.htbhf.claimant.message.MoneyUtils.convertPenceToPounds;
 
+/**
+ * Factory object for building message payloads for emails.
+ */
 public class MessagePayloadFactory {
 
     private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("dd MMM yyyy");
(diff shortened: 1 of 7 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactory.java
```diff
@@ -58,17 +61,37 @@ public static EmailMessagePayload buildSendNewCardSuccessEmailPayload(PaymentCyc
...
+        Map<String, Object> emailPersonalisation = createPaymentEmailPersonalisationMap(paymentCycle, entitlementNextMonth);
+        emailPersonalisation.put(EmailTemplateKey.MULTIPLE_CHILDREN.getTemplateKeyName(), multipleChildrenTurningFourInMonth);
+        return EmailMessagePayload.builder()
+                .claimId(paymentCycle.getClaim().getId())
+                .emailType(EmailType.CHILD_TURNS_FOUR)
+                .emailPersonalisation(emailPersonalisation)
+                .build();
+    }
+
+    private static Map<String, Object> createPaymentEmailPersonalisationMap(PaymentCycle paymentCycle, PaymentCycleVoucherEntitlement voucherEntitlement) {
```
gethyn1: Why has `voucherEntitlement` been extracted to a separate argument? (maybe you discussed this morning??)

mikeSmith08: Because we need to be emailing the about the entitlement that they will receive for their next payment, not their current payment (which is on the PaymentCycle).

PR #278, opened 2019-09-20
HTBHF-2288 Add in checking for child turning 4 in next cycle when new payment is created.

mikeSmith08:
Will remove the template id from the config in the next PR, but this adds the check and sending of the email to the payment message processor.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/MakePaymentMessageProcessor.java
```diff
@@ -30,18 +36,44 @@
...
     @Transactional(Transactional.TxType.REQUIRES_NEW)
     public MessageStatus processMessage(Message message) {
         MakePaymentMessageContext messageContext = messageContextLoader.loadMakePaymentContext(message);
-        paymentService.makePaymentForCycle(messageContext.getPaymentCycle(), messageContext.getCardAccountId());
-        EmailMessagePayload messagePayload = buildPaymentNotificationEmailPayload(messageContext.getPaymentCycle());
+        PaymentCycle paymentCycle = messageContext.getPaymentCycle();
+        paymentService.makePaymentForCycle(paymentCycle, messageContext.getCardAccountId());
+        EmailMessagePayload messagePayload = buildPaymentNotificationEmailPayload(paymentCycle);
         messageQueueClient.sendMessage(messagePayload, MessageType.SEND_EMAIL);
+        int numChildrenTurningFourInNextMonth = childDateOfBirthCalculator.getNumberOfChildrenTurningFourAffectingNextPayment(paymentCycle);
```
YetAnotherMatt: It's fine for now, but I'm pretty sure we're going to find we want to move the email decisions and generation into a separate component

mikeSmith08: Yeah, that's pretty obvious as this will grow but we'll leave it for now.

PR #279, opened 2019-09-22
HTBHF-2303 Replaced the single message schedule with one for each message type

YetAnotherMatt:
Currently we have a single message processor & schedule, so that all messages are processed one after another. With very few messages this is efficient and quick, but with more messages this becomes inefficient, and does not allow parallel processing on multiple instances. It also means that we cannot configure certain message types (such as sending email) to run only at specific times of the day.

Replace the single message processor schedule with one for each message type.


mikeSmith08:
You might want to move the story for this from Open into In Development. :-)

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -0,0 +1,90 @@
...
+    @SchedulerLock(
+            name = "Process ADDITIONAL_PREGNANCY_PAYMENT messages",
+            lockAtLeastForString = FIFTEEN_SECONDS,
+            lockAtMostForString = ONE_HOUR)
+    @NewRequestContextWithSessionId(sessionId = "MessageProcessor:ADDITIONAL_PREGNANCY_PAYMENT")
+    public void processAdditionalPregnancyPaymentMessages() {
+        messageProcessor.processMessagesOfType(MessageType.ADDITIONAL_PREGNANCY_PAYMENT);
+    }
+
+    @Scheduled(cron = "*/30 * 8-18 * * *")  // every 30 seconds between 8am and 6pm
```
mikeSmith08: I'm surprised the schedule isn't in config, is that going to be a subsequent PR?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -0,0 +1,90 @@
...
+import uk.gov.dhsc.htbhf.claimant.message.MessageProcessor;
+import uk.gov.dhsc.htbhf.claimant.message.MessageType;
+import uk.gov.dhsc.htbhf.requestcontext.aop.NewRequestContextWithSessionId;
+
+/**
+ * Component responsible for scheduling the processing of messages.
+ * Each MessageType has its own cron schedule - see https://stackoverflow.com/a/26147143 for examples.
+ * Each messageType also has its own minimum and maximum lock period. For a definition of the ISO-8601 duration format, see
+ * https://www.digi.com/resources/documentation/digidocs/90001437-13/reference/r_iso_8601_duration_format.htm.
+ * Note that the {@link MessageProcessor} also specifies a limit on the number of messages to process
```
mikeSmith08: We might also want to think about making this configurable on a message type basis too.

YetAnotherMatt: True

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
YetAnotherMatt: I wasn't sure if it needed to be in config. The question is, are we ever going to change this on a per-environment basis? If not, I'm not sure it needs to be in config.

mikeSmith08: Yep, that's fine, we can change when we need to.

PR #282, opened 2019-09-23
HTBHF-1757 Existing claims are updated to PENDING_EXPIRY if they beco…

dwybourn:
…me ineligible.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
@@ -24,6 +24,7 @@ import uk.gov.dhsc.htbhf.logging.event.FailureEvent;
 import java.time.LocalDateTime;
 import java.util.*;
 
+import static java.util.Collections.emptyList;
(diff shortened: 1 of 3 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimService.java
```diff
@@ -55,8 +56,14 @@ public ClaimResult createOrUpdateClaim(ClaimRequest claimRequest) {
 
         try {
             EligibilityAndEntitlementDecision decision = eligibilityAndEntitlementService.evaluateClaimant(claimRequest.getClaimant());
-            if (claimExistsAndIsEligible(decision)) {
+            if (claimExists(decision)) {
                 Claim claim = findClaim(decision.getExistingClaimId());
+
+                if (decisionIsNotEligible(decision)) {
```
YetAnotherMatt: This is the wrong place to invalidate a claim - we want to change the status to pending_expiry if the eligibility check for the payment_cycle reports ineligible, and leave the existing claim alone if a new, ineligible claim is received for that nino.

PR #287, opened 2019-09-24
HTBHF-1757 Moved logic of sending claim no longer eligible email into…

dwybourn:
… new class.


mikeSmith08:
Now I've had some time to think about this, I think we should be moving to a more OO design for this but having a set of Email objects that know how to build themselves. So in this case we'd have a ClaimNoLongerEligibleEmail object that took in two values (FirstName and LastName) and it knows how to build itself into a MessagePayload. You then don't need to have any large god-like objects and the top level manager class can simply construct those objects and ask them to build their own payload. I think that way forward would be better.

YetAnotherMatt:
> Now I've had some time to think about this, I think we should be moving to a more OO design for this but having a set of Email objects that know how to build themselves. So in this case we'd have a ClaimNoLongerEligibleEmail object that took in two values (FirstName and LastName) and it knows how to build itself into a MessagePayload. You then don't need to have any large god-like objects and the top level manager class can simply construct those objects and ask them to build their own payload. I think that way forward would be better.

That does mean that we couldn't mock them out - unless we had a factory for constructing them.

PR #289, opened 2019-09-24
HTBHF-1820 Moved update payment method into payment cycle class.

dwybourn:


mikeSmith08:
Any particular reason for this change or just a refactor? A comment on the purpose of the PR would have been useful.

PR #295, opened 2019-09-27
HTBHF-1658 Sending email notification when a child turns one in the next cycle.

dwybourn:
Sending email notification when a child turns one in the next cycle.


api/src/main/resources/application.yml
```diff
@@ -59,7 +59,7 @@ card:
 
 message-processor:
   # run every 30 seconds. See https://stackoverflow.com/a/26147143 for more examples
-  cron-schedule: "*/30 * * * * *"
+  cron-schedule: "*/2 * * * * *"
```
YetAnotherMatt: This probably wasn't intended as part of the commit

dwybourn: Reverted

PR #301, opened 2019-09-30
HTBHF-1820 Added integration test for child turning four.

dwybourn:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -239,6 +276,18 @@ private void assertChildTurnsOneEmailPersonalisationMap(PaymentCycle currentCycl
                 .contains(formatVoucherAmount(4)); // child is turning one in the next payment cycle, so one voucher per week going forward
     }
 
+    private void assertChildTurnsFourEmailPersonalisationMap(PaymentCycle currentCycle, Map childTurnsOnePersonalisationMap) {
```
YetAnotherMatt: I think that we'll end up sending a different email if we're stopping making payments. It might be better to make this test include payment for another child or pregnancy so that we don't have to change this test later?

dwybourn: Updated to include another child.

PR #310, opened 2019-10-02
HTBHF-1763 send new child from pregnancy email

gethyn1:
Send a "new child from pregnancy" email when a new child is a result of a pregnancy. This is determined by the existence of backdated vouchers on the payment cycle voucher entitlement.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/PaymentCycleVoucherEntitlementTest.java
```diff
@@ -148,4 +150,14 @@ void shouldReturnLastVoucherEntitlement() {
...
     }
 
+    @Test
+    void shouldReturnCorrectBackdatedVouchersValue() {
+        VoucherEntitlement voucherEntitlement = aValidVoucherEntitlement();
+        int backdatedVouchers = 10;
+
+        PaymentCycleVoucherEntitlement result = new PaymentCycleVoucherEntitlement(asList(voucherEntitlement), backdatedVouchers);
+
+        assertThat(result.getBackdatedVouchersValueInPence()).isEqualTo(backdatedVouchers * VOUCHER_VALUE_IN_PENCE);
```
mikeSmith08: This should be an explicit value rather than a calculation, it makes it harder to figure out if the test is correct or not.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/PaymentCycleVoucherEntitlementTestDataFactory.java
```diff
@@ -46,7 +46,7 @@ public static PaymentCycleVoucherEntitlement aPaymentCycleVoucherEntitlementMatc
 
     public static PaymentCycleVoucherEntitlement aPaymentCycleVoucherEntitlementWithBackdatedVouchersForYoungestChild(
             LocalDate startDate, List<LocalDate> childrensDob) {
-        LocalDate birthdate = childrensDob.stream().min(LocalDate::compareTo).get();
+        LocalDate birthdate = childrensDob.stream().max(LocalDate::compareTo).get();
```
mikeSmith08: Maybe this should be in its own method for clarity as it's obvious this can easily confuse.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entitlement/PaymentCycleVoucherEntitlementTest.java
gethyn1: Do you mean 1:
```
int expected = backdatedVouchers * VOUCHER_VALUE_IN_PENCE;
assertThat(result.getBackdatedVouchersValueInPence()).isEqualTo(expected);
```
.. or 2:
```
assertThat(result.getBackdatedVouchersValueInPence()).isEqualTo(3100);
```
?

mikeSmith08: I mean option 2, it's perfectly clear then what you're asserting.

YetAnotherMatt: I would argue that the meaning of the number is then less clear. It's just a magic number.

YetAnotherMatt: We're compromising by including a comment with the calculation

mikeSmith08: I don't agree with you there, I always feel it's better to be completely specific over what we are asserting and to not have your expected value be a calculation. Option 2 doesn't contain (or need) a comment either so I'm not sure what you mean there.

gethyn1: Updated

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/PaymentCycleVoucherEntitlementTestDataFactory.java
gethyn1: Updated

PR #313, opened 2019-10-03
HTBHF-2182 Setting a claim to expired when they have no vouchers.

dwybourn:
- Setting a claim to expired when they have no children under for an aren't pregnant (entitlement decision contains no vouchers).

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
@@ -13,6 +13,7 @@ import uk.gov.dhsc.htbhf.claimant.model.ClaimStatus;
 import uk.gov.dhsc.htbhf.claimant.model.eligibility.EligibilityAndEntitlementDecision;
 import uk.gov.dhsc.htbhf.claimant.repository.ClaimRepository;
 import uk.gov.dhsc.htbhf.claimant.service.EligibilityAndEntitlementService;
+import uk.gov.dhsc.htbhf.claimant.service.audit.EventAuditor;
 import uk.gov.dhsc.htbhf.claimant.service.payments.PaymentCycleService;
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -65,7 +69,10 @@ public MessageStatus processMessage(Message message) {
 
         paymentCycleService.updatePaymentCycle(currentPaymentCycle, decision);
 
-        if (decision.getEligibilityStatus() == ELIGIBLE) {
+        if (decision.getVoucherEntitlement().getTotalVoucherValueInPence() == 0) {
```
mikeSmith08: This should probably be checking the total number of vouchers rather than their total worth.

mikeSmith08: I also don't think this is right. Having tested this scenario, if there are no vouchers, then the decision that comes back from the eligibility check is INELIGIBLE, so I don't think this will work.

mikeSmith08:
It might have been better to wait to do this story until HTBHF-2185 was complete first so we're not treading on each other's toes here, plus the integration test that you need will already be in place then.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
dwybourn: Why would the decision being INELIGIBLE stop this check working?

dwybourn:
> It might have been better to wait to do this story until HTBHF-2185 was complete first so we're not treading on each other's toes here, plus the integration test that you need will already be in place then.

Having discussed with @YetAnotherMatt we're going to park this story until we've confirmed what we will get from the DWP api.

PR #314, opened 2019-10-03
HTBHF-2380 Add disabled integration test for an email being sent when…

mikeSmith08:
… there are no children returned for a claimant but they're still eligible.

Please have a good look at the test, the assertions on the state of the PaymentCycle are observations of what the system actually does rather than what it should do.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -217,33 +214,60 @@ void shouldSendNewChildEmailWhenPaymentCycleIncludesBackdatedVouchers() throws J
...
         verifyNoMoreInteractions(notificationClient);
     }
 
+    @Disabled("HTBHF-2185")
+    @Test
+    void shouldSendNoLongerEligibleEmailWhenEligibleWithNoChildrenOnFeedAndNotPregnant() throws JsonProcessingException, NotificationClientException {
+        // setup some claim variables
+        String cardAccountId = UUID.randomUUID().toString();
+
+        wiremockManager.stubSuccessfulEligibilityResponse(Collections.emptyList());
```
mikeSmith08: No card balance request is made is the eligibility status is determined to be ineligible.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -275,6 +299,17 @@ private void assertPaymentCycleIsFullyPaid(PaymentCycle paymentCycle, List<Local
         assertThat(payment.getPaymentAmountInPence()).isEqualTo(expectedVoucherEntitlement.getTotalVoucherValueInPence());
     }
 
+    private void assertPaymentCycleHasNoPayment(PaymentCycle paymentCycle, PaymentCycleVoucherEntitlement expectedVoucherEntitlement) {
+        assertThat(paymentCycle.getCycleStartDate()).isEqualTo(LocalDate.now());
+        assertThat(paymentCycle.getCycleEndDate()).isEqualTo(LocalDate.now().plusDays(27));
+        assertThat(paymentCycle.getChildrenDob()).isEmpty();
+        assertThat(paymentCycle.getVoucherEntitlement()).isEqualTo(expectedVoucherEntitlement);
+        assertThat(paymentCycle.getPaymentCycleStatus()).isEqualTo(PaymentCycleStatus.INELIGIBLE);
+        assertThat(paymentCycle.getCardBalanceInPence()).isNull();
```
mikeSmith08: This isn't set unless a payment is to be made, this doesn't feel correct, but its what the system currently does.

YetAnotherMatt: Until we have made a balance check we cannot populate `CardBalanceInPence`. As we're not making a payment, we don't make a balance check. What other value could `CardBalanceInPence` have?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -322,15 +370,15 @@ private void assertChildTurnsOneEmailPersonalisationMap(PaymentCycle currentCycl
                 .contains(formatVoucherAmount(4)); // child is turning one in the next payment cycle, so one voucher per week going forward
     }
 
-    private void assertChildTurnsFourEmailPersonalisationMap(PaymentCycle currentCycle, Map childTurnsOnePersonalisationMap) {
-        assertCommonEmailFields(currentCycle, childTurnsOnePersonalisationMap);
-        assertThat(childTurnsOnePersonalisationMap.get(CHILDREN_UNDER_1_PAYMENT.getTemplateKeyName())).asString()
+    private void assertChildTurnsFourEmailPersonalisationMap(PaymentCycle currentCycle, Map childTurnsFourPersonalisationMap) {
```
YetAnotherMatt: Trivial, but can we not just call it `personalisationMap`?

mikeSmith08: Yep, will do in the next PR.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
mikeSmith08: I think the point was that I thought we might always want to know what the balance was when a new PaymentCycle was created, rather than only retrieve it if we need to make a payment. Feels like it might be useful for MI purposes if nothing else.

PR #315, opened 2019-10-03
HTBHF-2330 Now deleting household to prevent errors when adding claim…

dwybourn:
Now deleting household to prevent errors when adding claimant with same details.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/creator/dwp/repository/UCHouseholdRepository.java
```diff
@@ -1,12 +1,38 @@
...
  * JPA gov.uk.dhsc.claimant.creator.repository for Universal Credit households (members of the household are persisted via the household).
  */
 public interface UCHouseholdRepository extends CrudRepository<UCHousehold, UUID> {
+
+    /**
+     * Return all households containing an adult with the given nino. These are ordered by fileImportNumber in descending order.
+     * @param nino The nino to check against
+     * @return A stream containing all households found
+     */
+    @Query("SELECT household FROM UCHousehold household INNER JOIN FETCH household.adults adult "
```
dwybourn: This code already existed in the DWP project.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/creator/ClaimantLoader.java
```diff
@@ -62,6 +63,9 @@ public void loadClaimantIntoDatabase() throws IOException {
     }
 
     private String createDWPHousehold(ClaimantInfo claimantInfo) {
+        Optional<UCHousehold> existingHousehold = ucHouseholdRepository.findHouseholdByAdultWithNino(claimantInfo.getNino());
```
mikeSmith08: For this test loader, I think it would be better to use `findAllHouseholdsByAdultWithNino` and delete all of the households found rather than presume that there is only ever going to be one and use the `findHouseholdByAdultWithNino` method.

YetAnotherMatt: agreed - delete all households with that NINO just in case

dwybourn: Updated to delete all households.

PR #319, opened 2019-10-04
HTBHF-2387 Added qualifyingBenefitStatus to EligibilityAndEntitlementDecision and PaymentCycle

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycle.java
```diff
@@ -45,6 +45,10 @@
     @Enumerated(EnumType.STRING)
     private EligibilityStatus eligibilityStatus;
 
+    @Column(name = "qualifying_benefit_status")
+    @Enumerated(EnumType.STRING)
+    private EligibilityStatus qualifyingBenefitStatus;
```
mikeSmith08: I'm not sure I understand why this isn't its own Enum, I thought they were going to diverge from the `eligibilityStatus`?

db/src/main/resources/db/migration/V1_040__add_qualifying_benefit_status_to_payment_cycle.sql
```diff
@@ -0,0 +1 @@
+alter table payment_cycle add column qualifying_benefit_status varchar;
```
mikeSmith08: What are you going to do for those claims already in the Production database?

YetAnotherMatt: Fair point. I'll add an update statement.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/PaymentCycle.java
YetAnotherMatt: It probably will diverge, but it currently has to be the same as the value that is returned by the eligibility service. I was talking about it with David on Friday, and we agreed that it will only ever have these 3 values, and that they can remain the same strings regardless of what enum we use:
```
    ELIGIBLE,
    INELIGIBLE,
    NO_MATCH,
```

mikeSmith08: Ok, thanks for clarifying, so these values will be the translated values, not the actual values sent back to us from DWP? I was wondering if it was a record of the qualifying benefit status then it might make sense to store that value instead.

YetAnotherMatt: I was thinking we'd translate 'confirmed' & 'not_confirmed' to 'ELIGIBLE' and 'INELIGIBLE'  in the eligibility service, as they have more meaning to us.

mikeSmith08: Yes, I get that, I was just wondering if storing the exact value we get back from DWP might be useful too.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/eligibility/QualifyingBenefitEligibilityStatus.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.dhsc.htbhf.claimant.model.eligibility;
+
+public enum QualifyingBenefitEligibilityStatus {
```
mikeSmith08: Worth some JavaDoc?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementService.java
```diff
@@ -114,4 +120,12 @@ private EligibilityStatus determineEligibilityStatus(EligibilityResponse respons
         return response.getEligibilityStatus();
     }
 
+    private QualifyingBenefitEligibilityStatus determineQualifyingBenefitStatus(EligibilityResponse response) {
```
mikeSmith08: Would it make sense for this to be in the enum itself called something like `fromEligibilityResponse`? Would certainly be easier to unit test then.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -165,7 +167,7 @@ void shouldReturnIneligibleForExistingClaimantWithoutVouchers() {
         EligibilityAndEntitlementDecision result =
                 eligibilityAndEntitlementService.evaluateExistingClaimant(claimant, cycleStartDate, previousCycle);
 
-        assertCorrectResult(result, INELIGIBLE, eligibilityResponse, voucherEntitlement);
+        assertCorrectResult(result, INELIGIBLE, CONFIRMED, eligibilityResponse, voucherEntitlement);
```
mikeSmith08: I am surprised not to see one of these tests have the status values of `INELIGIBLE, NOT_CONFIRMED`, is one of these tests wrong, or is there a missing test? Was expecting to see some tests with different `QualifyingBenefitEligibilityStatus` values being asserted.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleService.java
```diff
@@ -69,6 +70,7 @@ public PaymentCycle createAndSavePaymentCycleForEligibleClaim(Claim claim,
                 .cycleStartDate(cycleStartDate)
                 .cycleEndDate(cycleStartDate.plusDays(cycleDurationInDays - 1))
                 .eligibilityStatus(ELIGIBLE)
+                .qualifyingBenefitEligibilityStatus(CONFIRMED)
```
mikeSmith08: You should probably update the JavaDoc for this method too.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleServiceTest.java
```diff
@@ -160,13 +165,15 @@ void shouldUpdatePaymentCycleWithEligibilityAndEntitlementDecision() {
...
                 .eligibilityStatus(EligibilityStatus.INELIGIBLE)
+                .qualifyingBenefitEligibilityStatus(QualifyingBenefitEligibilityStatus.CONFIRMED)
                 .voucherEntitlement(voucherEntitlement)
                 .build();
 
         paymentCycleService.updatePaymentCycle(paymentCycle, decision);
 
         assertThat(paymentCycle.getId()).isEqualTo(paymentCycle.getId());
         assertThat(paymentCycle.getEligibilityStatus()).isEqualTo(decision.getEligibilityStatus());
+        assertThat(paymentCycle.getQualifyingBenefitEligibilityStatus()).isEqualTo(decision.getQualifyingBenefitEligibilityStatus());
```
mikeSmith08: I know we usually disagree on this type of assertion, but I do think this would be much clearer if it said:

`assertThat(paymentCycle.getQualifyingBenefitEligibilityStatus()).isEqualTo(QualifyingBenefitEligibilityStatus.CONFIRMED);`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
YetAnotherMatt: I have added a test for this, but also refactored the DuplicateClaimChecker. I'll put both in a separate pr.

PR #320, opened 2019-10-07
HTBHF-2387 Refactored DuplicateClaimChecker and added unit test

YetAnotherMatt:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -87,12 +88,34 @@ void shouldThrowExceptionWhenMultipleLiveClaimsExistForNino() {
...
+                LocalDate.now());
+    }
+
     @Test
     void shouldReturnEligibleWhenNotDuplicateAndEligibleWithVouchers() {
         EligibilityResponse eligibilityResponse = anEligibilityResponseWithStatus(ELIGIBLE);
         PaymentCycleVoucherEntitlement voucherEntitlement = aPaymentCycleVoucherEntitlementWithVouchers();
         given(client.checkEligibility(any())).willReturn(eligibilityResponse);
-        given(duplicateClaimChecker.checkForDuplicateClaimsFromHousehold(any())).willReturn(ELIGIBLE);
+        given(duplicateClaimChecker.liveClaimExistsForHousehold(any())).willReturn(false);
```
mikeSmith08: I think there is another missing test here, I don’t see any scenarios where `duplicateClaimChecker.liveClaimExistsForHousehold` returns `true`. Looks like there should be a test where the final `EligibiltyStatus` should be `DUPLICATE` when calling `evaluateClaimant` on `EligibilityAndEntitlementService`.

YetAnotherMatt: Good spot - we actually had a small bug in the code. Have refactored now.

PR #323, opened 2019-10-07
HTBHF-2413 Make sure that a voucher entitlement is only returned …

mikeSmith08:
…against the entitlement when claimant is determined to be ELIGIBLE.

This problem was highlighted when trying to assert that there was no voucher entitlement for a claimant who was INELIGIBLE in the new integration tests, but vouchers being returned and stored on the PaymentCycle.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementService.java
```diff
@@ -97,17 +97,30 @@ private EligibilityAndEntitlementDecision buildDecision(EligibilityResponse elig
                                                             PaymentCycleVoucherEntitlement entitlement,
                                                             UUID existingClaimId) {
         EligibilityStatus eligibilityStatus = determineEligibilityStatus(eligibilityResponse, entitlement);
+        PaymentCycleVoucherEntitlement voucherEntitlement = determinePaymentCycleVoucherEntitlementFromStatus(entitlement, eligibilityStatus);
```
YetAnotherMatt: Could we move this to before `EligibilityStatus eligibilityStatus = determineEligibilityStatus(eligibilityResponse, entitlement);`?
That way we still include the voucher entitlement if the reason we decided they were ineligible is because that entitlement showed zero vouchers (i.e. the eligibility service said they were eligible, but we calculated zero vouchers)

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementService.java
```diff
@@ -97,17 +97,30 @@ private EligibilityAndEntitlementDecision buildDecision(EligibilityResponse elig
...
         return EligibilityAndEntitlementDecision.builder()
                 .eligibilityStatus(eligibilityStatus)
                 .qualifyingBenefitEligibilityStatus(QualifyingBenefitEligibilityStatus.fromEligibilityStatus(eligibilityResponse.getEligibilityStatus()))
-                .voucherEntitlement(entitlement)
```
dwybourn: Would it not be better to do the check earlier and not even calculate the entitlement if the response is not ELIGIBLE?

mikeSmith08: The only problem with this is that we'd need to check again later on when we decide that they are INELIGIBLE because they are entitled to no vouchers, so makes sense to do it at the end.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementService.java
mikeSmith08: Why would we want to store the voucher entitlement if they are deemed to be INELIGIBLE because the entitlement has no vouchers in it? If we were going to do that then we might as well store an empty voucher entitlement for all scenarios otherwise we'll be inconsistent.

YetAnotherMatt: My thinking was: to show our working out. If we're not going to do that, then  I think we need to make sure we're recording that they weren't entitled to any vouchers, sowe need to set totalVouchers and totalEntitlementAmountInPence to zero (we can do that in `PaymentCycle.applyVoucherEntitlement` if the voucher is null).

YetAnotherMatt: As discussed, probably the best way to handle this is to add an `IneligibilityReason` enum & field. So this PR goes in as is, and we raise a story to store the reason for ineligibility.

mikeSmith08: Yep, that's a good plan.

PR #324, opened 2019-10-07
HTBHF-2412 Added report claim message processor.

dwybourn:
- Added report claim message processor which currently doesn't do anything and no messages of REPORT_CLAIM are created yet.
- Future PRs will create a REPORT_CLAIM message from the ClaimService and the processor will be updated to report the claim.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/ReportClaimMessageProcessor.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Responsible for reporting new and updated claims for MI reporting purposes.
+ * Fetches location data for a claim if it doesn't already exist, then reports the claim.
+ */
+public class ReportClaimMessageProcessor implements MessageTypeProcessor {
+
+    @Override
+    public MessageStatus processMessage(Message message) {
+        // TODO DW HTBHF-2412 Fetch location data for a claim and report the claim.
```
mikeSmith08: Shouldn't this be a different ticket number, HTBHF-2412 describes creating a no-op implementation only?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessagePayloadFactory.java
```diff
@@ -61,6 +58,12 @@ public static EmailMessagePayload buildEmailMessagePayload(PaymentCycle paymentC
                 .build();
     }
 
+    public static ReportClaimMessagePayload buildReportClaimMessagePayload(Claim claim) {
+        return ReportClaimMessagePayload.builder()
+                .claimId(claim.getId())
```
YetAnotherMatt: I think we'll also need to store the state of the claim (new/updated/rejected) as we can't assume the status of the claim will accurately reflect it later.

YetAnotherMatt: As discussed, we can add that in when we come to implement the story that needs it.

dwybourn: As discussed, this will be added with the additional fields that are required.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/ReportClaimMessageProcessor.java
dwybourn: Corrected to HTBHF-2009

PR #335, opened 2019-10-09
HTBHF-2418 Populating claim with postcode data from postcodes.io

dwybourn:
- Added PostcodeDataResponse object which represents the response from postcode.io.
- Populating claim with data from postcode.io.
- All postcode data fields are set to NOT_FOUND if postcodes.io returns a 404 for the given postcode.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/exception/PostcodesClientException.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.dhsc.htbhf.claimant.exception;
+
+/**
+ * Exception used when calling postcodes.io.
+ */
+public class PostcodesClientException extends RuntimeException {
```
mikeSmith08: Possibly `PostcodesIoClientException` unless you want to use it for other postcode clients (OS Places for example)?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporter.java
```diff
@@ -1,12 +1,68 @@
...
-        // TODO DW HTBHF-2418 Populate the claim with post code data and report claim to google analytics
+        if (claim.getPostcodeData() == null) {
+            PostcodeData postcodeData = getPostcodeData(claim);
+            claim.setPostcodeData(postcodeData);
+            claimRepository.save(claim);
+        }
+    }
+
+    private PostcodeData getPostcodeData(Claim claim) {
+        String postcode = claim.getClaimant().getAddress().getPostcode().replace(" ", "");
```
mikeSmith08: Is this worth being in its own method so you can explain why we need to do this? I presume we're doing this so we don't have a space in the built URL?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        assertThat(exception).isNotNull();
+        assertThat(claim.getPostcodeData()).isNull();
+        verify(restTemplate).getForObject(getExpectedPostcodeUrl(postcode), PostcodeDataResponse.class);
+        verifyZeroInteractions(claimRepository);
+    }
+
+    private String getExpectedPostcodeUrl(String postcode) {
+        String postcodeWithoutSpace = postcode.replace(" ", "");
+        return POSTCODES_IO_BASE_URI + POSTCODES_IO_PATH + postcodeWithoutSpace;
```
mikeSmith08: Won't this have two forward slashes in it, the first constant ends with one and the second starts with one?Is that right?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Test
+    void shouldNotGetPostcodeDataOrUpdateClaimWhenPostcodeDataExists() {
+        PostcodeData postcodeData = aPostcodeDataObjectForPostcode(VALID_POSTCODE);
+        Claim claim = aValidClaimBuilder().postcodeData(postcodeData).build();
+
+        miReporter.reportClaim(claim);
+
+        verifyZeroInteractions(restTemplate);
```
mikeSmith08: You can put multiple mocks into `verifyZeroInteractions` so `verifyZeroInteractions(restTemplate, claimRepository)`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
```diff
@@ -0,0 +1,126 @@
...
+    private MIReporter miReporter;
+
+    @BeforeEach
+    public void init() {
+        miReporter = new MIReporter(POSTCODES_IO_BASE_URI, restTemplate, claimRepository);
+    }
+
+    @Test
+    void shouldGetPostcodeDataAndSaveToClaim() {
+        Claim claim = aValidClaimBuilder().postcodeData(null).build();
```
mikeSmith08: Is it worth a test data factory method `aValidClaimWithPostcode` seeing as you have this repeated a number of times in this class.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
dwybourn: Updated to `verifyZeroInteractions(restTemplate, claimRepository)`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
dwybourn: Created aValidCalimWithPostcodeData method.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporterTest.java
dwybourn: It will in the test but the rest template doesn't care, I've updated the test value to not have a '/' at the end of the url.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/MIReporter.java
dwybourn: Moved into it's own method to be clearer.

PR #341, opened 2019-10-10
HTBHF-2028 Added regular payment amount to pregnancy email and update…

dwybourn:
- Added regular payment amount to new child from pregnancy email.
- Updated new child from pregnancy to use latest email template
- Refactored common code for creating an email personalisation map into EmailPayloadMessageFactory.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandlerTest.java
```diff
@@ -143,99 +153,76 @@ void shouldSendEmailForChildTurnsOneInFirstWeekOfNextCycle() {
...
+    private void assertEmailPayloadWhenChildTurnsFourOnFirstDayOfNextCycle(Map<String, Object> emailPersonalisation) {
+        assertThat(emailPersonalisation).contains(
                 entry("children_under_1_payment", "\n* £24.80 for children under 1"),
                 entry("children_under_4_payment", ""),
-                entry("multiple_children", false),
-                entry("next_payment_date", DATE_FORMATTER.format(nextPaymentDate)),
-                entry("regular_payment", "£24.80")
+                entry("multiple_children", false)
         );
+        assertThat(emailPersonalisation).containsAllEntriesOf(COMMON_EMAIL_MAP);
```
dwybourn: UpcomingBirthdayEmailHanlder is now adding to a common email map rather than creating one from scratch. Therefore the assertions now check for the presence of the added fields and the common map returned by the EmailMessagePayloadFactory.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
```diff
@@ -26,6 +26,28 @@
...
+        }
+        int totalAmount = numberOfVouchers * voucherAmountInPence;
+        return String.format(summaryTemplate, convertPenceToPounds(totalAmount));
+    }
+
+    public static String formatRegularPaymentAmount(int numberOfVouchers, int voucherAmountInPence) {
+        if (numberOfVouchers == 0) {
+            return "";
+        }
+        int totalAmount = numberOfVouchers * voucherAmountInPence;
```
mikeSmith08: Can we refactor the common code between this and `formatPaymentAmountSummary` above?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
```diff
@@ -73,4 +106,9 @@ private static String buildUnder4PaymentSummary(PaymentCycleVoucherEntitlement v
                 voucherEntitlement.getVouchersForChildrenBetweenOneAndFour(),
                 voucherEntitlement.getSingleVoucherValueInPence());
     }
+
+    private String getRegularPaymentAmountForNextCycle(PaymentCycleVoucherEntitlement voucherEntitlement) {
+        int totalVouchersForRegularPayment = voucherEntitlement.getLastVoucherEntitlementForCycle().getTotalVoucherEntitlement() * numberOfCalculationPeriods;
```
mikeSmith08: Why is this last voucher entitlement rather than first? Isn't the last more likely to be different than the first?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandler.java
```diff
@@ -77,25 +76,12 @@ private EmailMessagePayload buildChildTurnsAgeNotificationEmailPayload(PaymentCy
     }
 
     private Map<String, Object> createEmailPersonalisationMapForNextCycle(PaymentCycle paymentCycle, PaymentCycleVoucherEntitlement voucherEntitlement) {
-        Claimant claimant = paymentCycle.getClaim().getClaimant();
-        Map<String, Object> emailPersonalisation = new HashMap<>();
-        emailPersonalisation.put(FIRST_NAME.getTemplateKeyName(), claimant.getFirstName());
-        emailPersonalisation.put(LAST_NAME.getTemplateKeyName(), claimant.getLastName());
-        emailPersonalisation.put(PAYMENT_AMOUNT.getTemplateKeyName(), convertPenceToPounds(voucherEntitlement.getTotalVoucherValueInPence()));
-        emailPersonalisation.put(PREGNANCY_PAYMENT.getTemplateKeyName(), buildPregnancyPaymentAmountSummaryForNextCycle(voucherEntitlement));
+        Map<String, Object> emailPersonalisation = emailMessagePayloadFactory.createCommonEmailPersonalisationMap(paymentCycle, voucherEntitlement);
```
mikeSmith08: Can't you use `createPaymentEmailPersonalisationMap` here?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandlerTest.java
```diff
@@ -44,11 +43,14 @@
     private static final LocalDate TURNS_ONE_ON_DAY_OF_NEXT_PAYMENT_CYCLE = START_OF_NEXT_CYCLE.minusYears(1);
     private static final LocalDate TURNS_ONE_IN_FIRST_WEEK_OF_NEXT_PAYMENT_CYCLE = START_OF_NEXT_CYCLE.minusYears(1).plusDays(4);
     private static final Integer NUMBER_OF_CALCULATION_PERIODS = 4;
+    private static final Map<String, Object> COMMON_EMAIL_MAP = Map.of(FIRST_NAME.getTemplateKeyName(), VALID_FIRST_NAME);
```
mikeSmith08: Is it sufficient to have such a small map as this for these test? Might be but thought I'd ask the question.

dwybourn: I can't see how the size of the map matters, the key thing is that we test that we have the map and add to it.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
dwybourn: As discussed, this is what the regular payment going forward will be.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandlerTest.java
YetAnotherMatt: We're just testing that the contents of this map are included. For that purpose, a single entry is enough

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandler.java
YetAnotherMatt: now using createCommonEmailPersonalisationMap

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
YetAnotherMatt: Is this better?
```

    public static String formatPaymentAmountSummary(String summaryTemplate, int numberOfVouchers, int voucherAmountInPence) {
        return formatPaymentAmount(summaryTemplate, numberOfVouchers, voucherAmountInPence);
    }

    public static String formatRegularPaymentAmount(int numberOfVouchers, int voucherAmountInPence) {
        return formatPaymentAmount("%s", numberOfVouchers, voucherAmountInPence);
    }

    private static String formatPaymentAmount(String template, int numberOfVouchers, int voucherAmountInPence) {
        if (numberOfVouchers == 0) {
            return "";
        }
        int totalAmount = numberOfVouchers * voucherAmountInPence;
        return String.format(template, convertPenceToPounds(totalAmount));
    }
```

mikeSmith08: Yep, looks good.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/UpcomingBirthdayEmailHandler.java
mikeSmith08: My point was that if you use `createPaymentEmailPersonalisationMap` instead of `createCommonEmailPersonalisationMap` then it looks like you can get rid of the two lines below too (children under 1 payment and children under 4 payment).

YetAnotherMatt: Sadly not - those two lines do something different to the equivalent lines in `createPaymentEmailPersonalisationMap`. We did start out trying to re-use that method.

PR #343, opened 2019-10-10
HTBHF-2010 Added dates of birth of children and claim action to repor…

dwybourn:
- Added dates of birth of children and claim action to report claim message
- Now only creating report claim message for a new claim (other cases are handled in other stories)
- Added ClaimAction enum that states what happened to trigger a report claim message, e.g. new, updated, pending_expiry-> active, pending_expiry->expired, etc.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimAction.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimAction.java
@@ -1,0 +1,8 @@
+package uk.gov.dhsc.htbhf.claimant.reporting;
+
+/**
+ * Action that triggered a claim event.
+ */
+public enum ClaimAction {
+    NEW
+}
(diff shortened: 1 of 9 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimAction.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.dhsc.htbhf.claimant.reporting;
+
+/**
+ * Action that triggered a claim event.
+ */
+public enum ClaimAction {
```
mikeSmith08: Is there a reason for a new enum, or could we use `ClaimStatus` here?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimMessageSender.java
```diff
@@ -24,8 +28,8 @@
 
     private final MessageQueueClient messageQueueClient;
 
-    public void sendReportClaimMessage(Claim claim) {
-        ReportClaimMessagePayload payload = buildReportClaimMessagePayload(claim);
+    public void sendReportClaimMessage(Claim claim, List<LocalDate> datesOfBirthOfChildren, ClaimAction claimAction) {
```
mikeSmith08: Would this method be less clunky if it was simply called `reportClaim`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimAction.java
dwybourn: It's an action rather than a status, so for a new claim it's the same as the status, but other values will include transitions such as 'pending_expiry to expired' or 'active to expired'.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/ClaimMessageSender.java
dwybourn: It's not reporting the claim, just creating the message on the queue. Reporting the claim will be inside the MIReporter class.

PR #344, opened 2019-10-11
HTBHF-2381 Implement check for claimant no longer being eligible for …

mikeSmith08:
…the scheme by not being pregnant and DWP reporting no children. Claim is set to EXPIRED and claimant emailed to say they're off the scheme.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -265,7 +264,7 @@ void shouldSendNoLongerEligibleEmailWhenEligibleWithNoChildrenOnFeedAndNotPregna
         PaymentCycle newCycle = repositoryMediator.getCurrentPaymentCycleForClaim(claim);
         assertPaymentCycleWithNoPayment(newCycle, currentPaymentCycleChildrenDobs);
 
-        assertStatusOnClaim(claim, ClaimStatus.PENDING_EXPIRY);
+        assertStatusOnClaim(claim, ClaimStatus.EXPIRED);
```
mikeSmith08: We got the status wrong in the integration test, this scenario should be setting the status to EXPIRED.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -407,8 +406,7 @@ void shouldTestClaimRemainingPendingExpiryIfLessThan16Weeks(List<LocalDate> prev
                 Arguments.of(SINGLE_THREE_YEAR_OLD, SINGLE_THREE_YEAR_OLD, DUE_DATE_IN_4_MONTHS),
                 Arguments.of(NO_CHILDREN, SINGLE_THREE_YEAR_OLD, DUE_DATE_IN_4_MONTHS),
                 Arguments.of(SINGLE_THREE_YEAR_OLD, NO_CHILDREN, DUE_DATE_IN_4_MONTHS),
-                Arguments.of(NO_CHILDREN, NO_CHILDREN, DUE_DATE_IN_4_MONTHS),
-                Arguments.of(SINGLE_THREE_YEAR_OLD, NO_CHILDREN, NOT_PREGNANT)
```
mikeSmith08: This integration test scenario was invalid for HTBHF-1757 - if the claimant has no children in the current cycle and is not pregnant then that scenario is the same as that for this story, hence they should be set to EXPIRED and sent an email to say they are no longer on the service. Have removed this scenario from that parameterised test.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -57,36 +57,52 @@ public MessageType supportsMessageType() {
...
+            handleNoLongerEligibleForScheme(claim);
         } else {
-            handleNonEligibleClaim(currentPaymentCycle);
+            handleNonEligibleQualifyingBenefitStatusClaim(claim);
         }
+    }
 
-        return COMPLETED;
+    private boolean claimantIsNotPregnant(Claimant claimant) {
+        return claimant.getExpectedDeliveryDate() == null;
```
dwybourn: Needs to be updated to take into account claims with pregnancy but are too far in the past.

mikeSmith08: Done, we now use PregnancyEntitlementCalculator to do that check.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -57,36 +59,56 @@ public MessageType supportsMessageType() {
...
+    }
 
+    private void handleDecision(Claim claim, PaymentCycle currentPaymentCycle, EligibilityAndEntitlementDecision decision) {
         if (decision.getEligibilityStatus() == ELIGIBLE) {
             createMakePaymentMessage(currentPaymentCycle);
+        } else if (decision.noChildrenPresentInCurrentCycle() && claimantIsNotPregnant(claim, currentPaymentCycle)) {
+            handleNoLongerEligibleForScheme(claim);
         } else {
-            handleNonEligibleClaim(currentPaymentCycle);
+            handleNonEligibleQualifyingBenefitStatusClaim(claim);
```
YetAnotherMatt: how about `handleLossOfQualifyingBenefitStatus`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
YetAnotherMatt: Wasn't there a difference between sudden loss of a child (i.e. last cycle there was a child who would still be under 4) and the last child turning 4?

mikeSmith08: At the moment, this scenario results in claim being set to EXPIRED and the claimant being sent an email to say that they are no longer eligible because they have no children and are not pregnant, so yes, it is different to their child turning 4, but this scenario doesn't fit in with these tests and hence why I removed it, does that make sense?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
mikeSmith08: Yes, that's good have changed.

PR #347, opened 2019-10-14
HTBHF-2010 Creating event properties and custom dimensions for google…

dwybourn:
- First part of constructing the payload for report a new claim to google analytics. This consists of event properties, custom dimensions and custom metrics.
- Creating event properties and custom dimensions for google analytics.
- Adding custom metrics will be added in a future PR as it requires a refactor of some existing code and would make this PR quite large.
- List of event properties, custom metrics and dimensions are in the jira ticket.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Determines the {@link uk.gov.dhsc.htbhf.claimant.reporting.payload.ClaimantCategory} for a given claimant and their children's date of birth.
+     *
+     * @param claimant               the claimant to determine the category of
+     * @param datesOfBirthOfChildren the dates of birth of the claimant's children
+     * @param timestamp the timestamp to use when checking a claimant/child's age
+     * @return claimant's determined category
+     */
+    public String determineClaimantCategory(Claimant claimant, List<LocalDate> datesOfBirthOfChildren, LocalDateTime timestamp) {
```
mikeSmith08: Why does this take a LocalDateTime rather than just a LocalDate? Don't think we'd ever need the time element so would probably be better with just the LocalDate.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+import static uk.gov.dhsc.htbhf.claimant.reporting.payload.ClaimantCategory.NOT_PREGNANT_WITH_CHILDREN;
+import static uk.gov.dhsc.htbhf.claimant.reporting.payload.ClaimantCategory.PREGNANT_AND_UNDER_16;
+import static uk.gov.dhsc.htbhf.claimant.reporting.payload.ClaimantCategory.PREGNANT_WITH_CHILDREN;
+import static uk.gov.dhsc.htbhf.claimant.reporting.payload.ClaimantCategory.PREGNANT_WITH_NO_CHILDREN;
+
+@Component
+@RequiredArgsConstructor
+public class ClaimantCategoryCalculator {
```
mikeSmith08: Looking at this, I was wondering if it could mostly be encapsulated within an Enum which would simplify this greatly. From looking at it we would need initially calculate three values (ageInYears, isClaimantPregnant and hasChildren) and simply lookup the enum value based from those answers (which would be in the enum). What do you think?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,108 @@
...
+                .build();
+        given(pregnancyEntitlementCalculator.isEntitledToVoucher(any(), any())).willReturn(true);
+
+        String claimantCategory = claimantCategoryCalculator.determineClaimantCategory(claimant, emptyList(), LocalDateTime.now());
+
+        LocalDate now = LocalDate.now();
+        assertThat(claimantCategory).isEqualTo("Pregnant and under 16");
+        ArgumentCaptor<LocalDate> argumentCaptor = ArgumentCaptor.forClass(LocalDate.class);
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(eq(expectedDeliveryDate), argumentCaptor.capture());
+        assertThat(argumentCaptor.getValue()).isAfterOrEqualTo(now);
```
mikeSmith08: As this value is just going to be LocalDate.now(), why do you need this?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,108 @@
...
+                .build();
+        given(pregnancyEntitlementCalculator.isEntitledToVoucher(any(), any())).willReturn(false);
+
+        String claimantCategory = claimantCategoryCalculator.determineClaimantCategory(claimant, datesOfBirthOfChildren, LocalDateTime.now());
+
+        LocalDate now = LocalDate.now();
+        assertThat(claimantCategory).isEqualTo("Not pregnant with passported children");
+        ArgumentCaptor<LocalDate> argumentCaptor = ArgumentCaptor.forClass(LocalDate.class);
+        verify(pregnancyEntitlementCalculator, times(3)).isEntitledToVoucher(eq(expectedDeliveryDate), argumentCaptor.capture());
+        assertThat(argumentCaptor.getValue()).isAfterOrEqualTo(now);
```
mikeSmith08: You should refactor out all these duplicate calls to check the date (passing in the number of times of course).

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.ClaimantTestDataFactory.aValidClaimantBuilder;
+
+@ExtendWith(MockitoExtension.class)
+class ClaimantCategoryCalculatorTest {
```
mikeSmith08: Feels like all these tests might be able to be a parameterised test, I'm not 100% sure but there should be at least a fair bit of common code that could be refactored if not.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+                .datesOfBirthOfChildren(datesOfBirthOfChildren)
+                .timestamp(timeStamp)
+                .build();
+        given(claimantCategoryCalculator.determineClaimantCategory(any(), any(), any())).willReturn(claimantCategory);
+
+        Map<String, Object> reportProperties = reportPropertiesFactory.createReportPropertiesForClaimEvent(context);
+
+        // Queue time is number of milliseconds since now and the timestamp value. Therefore a timestamp one second ago should have a queue time >= 1000
+        assertThat((Long) reportProperties.get("queueTime")).isGreaterThanOrEqualTo(1000);
+        assertThat(reportProperties).contains(
```
mikeSmith08: This should be `containsExactly` otherwise anything additional in the Map will not cause the test to fail.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+                .claim(claim)
+                .datesOfBirthOfChildren(datesOfBirthOfChildren)
+                .timestamp(timeStamp)
+                .build();
+        given(claimantCategoryCalculator.determineClaimantCategory(any(), any(), any())).willReturn(claimantCategory);
+
+        Map<String, Object> reportProperties = reportPropertiesFactory.createReportPropertiesForClaimEvent(context);
+
+        // Queue time is number of milliseconds since now and the timestamp value. Therefore a timestamp one second ago should have a queue time >= 1000
+        assertThat((Long) reportProperties.get("queueTime")).isGreaterThanOrEqualTo(1000);
```
mikeSmith08: Isn't this a bit dangerous? Isn't it sufficient to say that it's greater than or equal to 1? Why the choice of 1000?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
YetAnotherMatt: I suggest this method returns a ClaimantCategory, and always returns a value. The logic can be simplified a bit:
```
if (isClaimantPregnant(claimant)) {
	if (claimantAgeInYears < 16) {
		return PREGNANT_AND_UNDER_16;
	}
	if (hasChildren(datesOfBirthOfChildren)) {
	   return PREGNANT_WITH_CHILDREN;
	}
	return PREGNANT_WITH_NO_CHILDREN;
}
if (hasChildren(datesOfBirthOfChildren)) {
   return NOT_PREGNANT_WITH_CHILDREN;
}
// this could happen once a claimant has expired
return NOT_PREGNANT_NO_CHILDREN;
```

YetAnotherMatt: Actually, we're also missing this category: Pregnant under 18/20FTE
I know we can't determine the under 20 FTE part, but we can determine under 18

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
```diff
@@ -0,0 +1,60 @@
...
+        } else if (!isClaimantPregnant(claimant) && hasChildren(datesOfBirthOfChildren)) {
+            return NOT_PREGNANT_WITH_CHILDREN.getDescription();
+        }
+
+        // TODO DW HTBHF-2010 handle case where claimant doesn't match any category.
+        return null;
+    }
+
+    private boolean hasChildren(List<LocalDate> datesOfBirthOfChildren) {
+        return datesOfBirthOfChildren != null && !datesOfBirthOfChildren.isEmpty();
```
YetAnotherMatt: There's an `isEmpty(datesOfBirthOfChildren)` commons method for this

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    private int getClaimantAgeInYears(Claimant claimant, LocalDateTime timestamp) {
+        return Period.between(claimant.getDateOfBirth(), timestamp.toLocalDate()).getYears();
+    }
+
+    private boolean isClaimantPregnant(Claimant claimant) {
```
YetAnotherMatt: This should use the same timestamp as other methods instead of LocalDate.now() (though as Mike says we should be using LocalDate not LocalDateTime

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
```diff
@@ -0,0 +1,65 @@
...
+            return PREGNANT_WITH_NO_CHILDREN;
+        }
+        if (hasChildren(datesOfBirthOfChildren, atDate)) {
+            return NOT_PREGNANT_WITH_CHILDREN;
+        }
+        // this could happen once a claim has expired
+        return NOT_PREGNANT_WITH_NO_CHILDREN;
+    }
+
+    private boolean hasChildren(List<LocalDate> datesOfBirthOfChildren, LocalDate atDate) {
```
mikeSmith08: Feels like this should go in `ChildDateOfBirthCalculator` its already doing a similar calculation for a PaymentCycle so might be good to combine with the method in there.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
```diff
@@ -0,0 +1,105 @@
...
+    private Map<String, Object> createCustomMetricMap(ReportClaimMessageContext context) {
+        Map<String, Object> customMetrics = new TreeMap<>();
+        Claimant claimant = context.getClaim().getClaimant();
+        customMetrics.put(CLAIMANT_AGE.getFieldName(), Period.between(claimant.getDateOfBirth(), LocalDate.now()).getYears());
+        long childrenUnder4 = context.getDatesOfBirthOfChildren().stream().filter(dob -> dob.isAfter(LocalDate.now().minusYears(4))).count();
+        long childrenUnder1 = context.getDatesOfBirthOfChildren().stream().filter(dob -> dob.isAfter(LocalDate.now().minusYears(1))).count();
+        customMetrics.put(CHILDREN_UNDER_ONE.getFieldName(), childrenUnder1);
+        customMetrics.put(CHILDREN_BETWEEN_ONE_AND_FOUR.getFieldName(), childrenUnder4 - childrenUnder1);
+        LocalDate expectedDeliveryDate = claimant.getExpectedDeliveryDate();
+        if (expectedDeliveryDate == null || expectedDeliveryDate.isBefore(LocalDate.now())) {
```
mikeSmith08: Should this be using the same component that determines if a delivery date is considered pregnant or not?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
YetAnotherMatt: The map does contain a lot of other values

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+        assertThat(reportProperties).contains(
+                entry("cm1", String.valueOf(childrenUnderOne)),
+                entry("cm2", String.valueOf(childrenOneToFour))
+        );
+    }
+
+    static Stream<Arguments> childrensDatesOfBirth() {
+        return Stream.of(
+                Arguments.of(List.of(LocalDate.now().minusDays(1), LocalDate.now().minusMonths(11)), 2, 0),
```
mikeSmith08: Could we use constants for these dates of birth so that its easier to see the specific meaning of each date as we've done in other tests?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
dwybourn: This and other methods have been updated to use a LocalDate.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
dwybourn: Left over from when it wasn't taking in the timestamp. It has now been updated to assert the timestamp.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
dwybourn: Agreed. Will add in next PR along with moving some other refactors.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
dwybourn: Will do in next PR along with some other refactors.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculator.java
dwybourn: Added.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
dwybourn: Added a variable and used TimeUnit.toMilliseconds to make this clearer.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
dwybourn: Code has been refactored so there are no longer multiple calls.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
dwybourn: Done.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
dwybourn: We also can't do that as we don't have an exact value for queueTime so it needs a separate assertion.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
dwybourn: Done.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        ClaimantCategory claimantCategory = claimantCategoryCalculator.determineClaimantCategory(claimant, emptyList(), atDate);
+
+        assertThat(claimantCategory).isEqualTo(expectedCategory);
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(expectedDeliveryDate, atDate);
+    }
+
+    @Test
+    void shouldMatchClaimantToPregnantWithChildren() {
+        LocalDate expectedDeliveryDate = LocalDate.now().plusMonths(2);
```
mikeSmith08: Can we use `TestConstants.EXPECTED_DELIVERY_DATE_IN_TWO_MONTHS` here please?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        assertThat(claimantCategory).isEqualTo(PREGNANT_WITH_CHILDREN);
+        ArgumentCaptor<LocalDate> argumentCaptor = ArgumentCaptor.forClass(LocalDate.class);
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(eq(expectedDeliveryDate), argumentCaptor.capture());
+        assertThat(argumentCaptor.getValue()).isAfterOrEqualTo(now);
+    }
+
+    @Test
+    void shouldMatchClaimantToNotPregnantWithChildren() {
+        LocalDate expectedDeliveryDate = null;
```
mikeSmith08: Minor - can we use `TestConstants.NOT_PREGNANT` for clarity?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        assertThat(claimantCategory).isEqualTo(NOT_PREGNANT_WITH_NO_CHILDREN);
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(expectedDeliveryDate, now);
+    }
+
+    @Test
+    void shouldMatchClaimantToNotPregnantWithNoChildrenWhenChildrenOver4() {
+        LocalDate expectedDeliveryDate = null;
+        LocalDate now = LocalDate.now();
+        List<LocalDate> datesOfBirthOfChildren = singletonList(LocalDate.now().minusYears(4));
```
mikeSmith08: Is this specifically testing the 4 year boundary case, if so, it should be explicit in the test name, otherwise I'd make their DOB a bit older so its definitely over 4 and not just 4.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimantCategoryCalculatorTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Claimant claimant = aValidClaimantBuilder()
+                .expectedDeliveryDate(expectedDeliveryDate)
+                .dateOfBirth(LocalDate.now().minusYears(20))
```
mikeSmith08: Might be worth a constant for `20_YEAR_OLD` seeing as you use it a lot.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,217 @@
...
+        ReportClaimMessageContext context = aReportClaimMessageContext(timestamp, datesOfBirthOfChildren, EXPECTED_DELIVERY_DATE_IN_TWO_MONTHS);
+        given(claimantCategoryCalculator.determineClaimantCategory(any(), any(), any())).willReturn(CLAIMANT_CATEGORY);
+
+        Map<String, String> reportProperties = reportPropertiesFactory.createReportPropertiesForClaimEvent(context);
+
+        // Queue time is number of milliseconds since now and the timestamp value. Therefore a timestamp one second ago should have a queue time >= 1000
+        Long queueTime = Long.parseLong(reportProperties.get("qt"));
+        assertThat(queueTime).isGreaterThanOrEqualTo(TimeUnit.SECONDS.toMillis(secondsSinceEvent));
+        long maxQueueTime = ChronoUnit.MILLIS.between(timestamp, LocalDateTime.now());
+        assertThat(queueTime).isLessThanOrEqualTo(maxQueueTime);
```
mikeSmith08: Can we use `assertThat().isBetween()` here instead of the two separate assertions?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,217 @@
...
+        // queue times longer than 4 hours may cause the event to not be registered:
+        // https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#qt
+        Long queueTime = Long.parseLong(reportProperties.get("qt"));
+        long maxQueueTime = 1000 * 60 * 60 * 4;
+        assertThat(queueTime).isLessThanOrEqualTo(maxQueueTime);
+    }
+
+    @Test
+    void shouldIncludeMetricsForZeroChildren() {
+        ReportClaimMessageContext context = aReportClaimMessageContext(LocalDateTime.now(), emptyList(), null);
```
mikeSmith08: Constants would help legibility here too (plus below too).

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+
+    @ParameterizedTest(name = "{1} weeks pregnant given {0} since conception")
+    @CsvSource({
+            "P10W, 10",
+            "P20W1D, 20",
+            "P36W6D, 36"
+    })
+    void shouldIncludeMetricsForPregnantClaimant(String timeSinceConception, Long weeksPregnant) {
+        LocalDate expectedDeliveryDate = LocalDate.now().minus(Period.parse(timeSinceConception)).plusMonths(9);
```
mikeSmith08: Should we be thinking of a pregnancy as 9 months or 40 weeks?

PR #350, opened 2019-10-15
HTBHF-2182 Make sure that the default Claimant is pregnant as this is…

mikeSmith08:
… the majority of its usage. Also make sure that the expected delivery date is set on any PaymentCycle built from the Claimant provided in the test data factory.

The existing PaymentCycles were building a voucher entitlement that had vouchers for pregnancy but the Claimant being used didn't have an expected delivery date. This PR attempts to fix that and make sure that the expected delivery date is what's used on the PaymentCycle too.

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
@@ -35,6 +35,8 @@ public class TestConstants {
 
     public static final LocalDate MAGGIE_DOB = LocalDate.now().minusMonths(6);
     public static final LocalDate LISA_DOB = LocalDate.now().minusMonths(24);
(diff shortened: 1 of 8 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -35,6 +35,8 @@
 
     public static final LocalDate MAGGIE_DOB = LocalDate.now().minusMonths(6);
     public static final LocalDate LISA_DOB = LocalDate.now().minusMonths(24);
+    public static final LocalDate EXPECTED_DELIVERY_DATE = LocalDate.now().plusMonths(2);
```
YetAnotherMatt: How about appending `_IN_2_MONTHS` to the name?

mikeSmith08: Yep, will do in the next PR.

PR #355, opened 2019-10-17
HTBHF-2480 Made claimStatus private to claim, added updateClaimStatus method

YetAnotherMatt:


api/build.gradle
```diff
@@ -71,7 +71,7 @@ dependencies {
     implementation "net.javacrumbs.shedlock:shedlock-spring:2.4.0"
     implementation "net.javacrumbs.shedlock:shedlock-provider-jdbc-template:2.4.0"
     implementation "com.vladmihalcea:hibernate-types-52:2.4.3"
-    implementation "org.javers:javers-spring-boot-starter-sql:5.5.0:"
+    implementation "org.javers:javers-spring-boot-starter-sql:5.7.7:"
```
YetAnotherMatt: An unrelated change. I got confused by the order of changes in Javers so updated it to while I was figuring it out. I decided it did no harm to be using the latest version.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claim.java
```diff
@@ -66,4 +70,18 @@
...
     private Claimant claimant;
+
+    public void updateClaimStatus(ClaimStatus newStatus) {
+        if (claimStatus != newStatus) {
+            claimStatusTimestamp = LocalDateTime.now();
+        }
+        claimStatus = newStatus;
+    }
+
+    public void updateEligibilityStatus(EligibilityStatus newStatus) {
```
YetAnotherMatt: We're not actually using this method yet (i.e. we're never updating eligibilityStatus).

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -88,7 +88,7 @@ public void processPaymentMessages() {
         messageProcessor.processMessagesOfType(MessageType.MAKE_PAYMENT);
     }
 
-    @Scheduled(cron = EVERY_30_SECONDS)
+    @Scheduled(cron = "05/35 * * * * *") // every 30 seconds, offset from other schedules by 5 seconds to prevent clashes with CREATE_NEW_CARD for new claims
```
YetAnotherMatt: Again, unrelated, but I saw that for new claims the CREATE_NEW_CARD and REPORT_CLAIM messages were always clashing and causing one of them to fail with a hibernate concurrent update exception. This offsets reporting new claims by 5 seconds to prevent that.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimTest.java
```diff
@@ -86,4 +92,58 @@ void shouldReturnTheSameIdIfOneIsSet() {
...
+        LocalDateTime originalTimestamp = LocalDateTime.now().minusHours(1);
+        Claim claim = Claim.builder()
+                .claimStatus(ClaimStatus.NEW)
+                .claimStatusTimestamp(originalTimestamp)
+                .build();
+        LocalDateTime now = LocalDateTime.now();
+
+        claim.updateClaimStatus(ClaimStatus.ACTIVE);
+
+        Assertions.assertThat(claim.getClaimStatusTimestamp()).isAfterOrEqualTo(now);
```
mikeSmith08: This should just be `assertThat`, don't need/want the `Assertions` part in front of it.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimTest.java
```diff
@@ -86,4 +92,58 @@ void shouldReturnTheSameIdIfOneIsSet() {
         //Then
         Assertions.assertThat(id).isEqualTo(claim.getId());
     }
+
+    @Test
+    void shouldUpdateClaimStatusAndTimestamp() {
+        LocalDateTime originalTimestamp = LocalDateTime.now().minusHours(1);
```
mikeSmith08: I think after the usage of constants for times in other tests, it would be good to create a constant in this test class called `ONE_HOUR_AGO` to use in all these tests, it'll be really clear then.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/ClaimTest.java
YetAnotherMatt: Unfortunately, there are other assertions in this class that make `Assertions.` necessary.

YetAnotherMatt: e.g. `assertThat(violations).hasSingleConstraintViolation`

PR #356, opened 2019-10-18
HTBHF-2182 Add check to DetermineEntitlementMessageProcessor for expiring…

mikeSmith08:
… claims, enable the corresponding integration test and add missing unit tests for this processor.

 - Rewritten the handleDecision method in DetermineEntitlementMessageProcessor to match the logic from the whiteboard session, some scenarios haven't been implemented yet so have corresponding TODOs.
 - Enabled the integration test for this scenario
 - The unit tests for DetermineEntitlementMessageProcessor were considerably lacking and a review of the code coverage for the new code highlighted this so I had to create a number of extra tests to cover all permutations.
 - Added method to QualifyingBenefitStatus to determine if the status is not eligible or not (specifically not using the term ineligible)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -146,16 +224,162 @@ void shouldUpdateClaimWhenClaimantIsNotEligible(EligibilityStatus eligibilitySta
...
+        LocalDate currentPaymentCycleStartDate = context.getCurrentPaymentCycle().getCycleStartDate();
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(NOT_PREGNANT, currentPaymentCycleStartDate);
+        verify(childDateOfBirthCalculator).hadChildrenUnder4AtStartOfPaymentCycle(context.getPreviousPaymentCycle());
+        verify(childDateOfBirthCalculator).hadChildrenUnderFourAtGivenDate(TWO_CHILDREN, currentPaymentCycleStartDate);
+        verifyZeroInteractions(childDateOfBirthCalculator);
+    }
+
+    @ParameterizedTest(name = "Qualifying benefit status={0}")
+    @ValueSource(strings = {"CONFIRMED", "NOT_CONFIRMED"})
+    void shouldExpireClaimWhenClaimantWasPregnantButNoLongerWithChildrenInPreviousCycle(
```
mikeSmith08: I wanted to combine this test and the next test but they would need conditional logic on the mocks so decided to separate them out into two tests, if anyone can see a good way, please let me know.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -146,16 +224,162 @@ void shouldUpdateClaimWhenClaimantIsNotEligible(EligibilityStatus eligibilitySta
...
+        //Then
+        runCommonVerifications(context, decision, message, messageStatus, ClaimStatus.EXPIRED);
+        verify(childDateOfBirthCalculator).hadChildrenUnder4AtStartOfPaymentCycle(context.getPreviousPaymentCycle());
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(NOT_PREGNANT, context.getCurrentPaymentCycle().getCycleStartDate());
+        verifyZeroInteractions(determineEntitlementNotificationHandler);
+    }
+
+    @ParameterizedTest(name = "Qualifying benefit status={0}")
+    @ValueSource(strings = {"CONFIRMED", "NOT_CONFIRMED"})
+    void shouldExpireClaimWhenClaimantWasPregnantButNoLongerWithNoChildrenInPreviousCycle(
```
mikeSmith08: I'm also unsure whether this and the previous test are correct. If the claimant has no children, was previously pregnant and is now no longer pregnant, should we expire them with no email? I'm double checking with Karen...

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -69,26 +72,48 @@ public MessageStatus processMessage(Message message) {
...
+    }
+
+    private boolean shouldExpireClaim(EligibilityAndEntitlementDecision decision, PaymentCycle previousPaymentCycle, PaymentCycle currentPaymentCycle) {
+        if (decision.childrenPresent() || claimantIsPregnantInCycle(currentPaymentCycle)) {
+            return false;
         }
+        if (childrenExistedInPreviousCycleAndNowOver4(previousPaymentCycle, currentPaymentCycle)) {
+            return true;
+        }
+        return claimantIsPregnantInCycle(previousPaymentCycle) && !claimantIsPregnantInCycle(currentPaymentCycle);
```
dwybourn: Do we need to check if the claimant was pregnant previously?
I think we're missing the case where the children go off the feed without turning four. (loss of child).

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityAndEntitlementTestDataFactory.java
```diff
@@ -1,38 +1,51 @@
...
+                                                                                   List<LocalDate> childrenDobs) {
+        EligibilityAndEntitlementDecision.EligibilityAndEntitlementDecisionBuilder builder = aValidDecisionBuilder().eligibilityStatus(eligibilityStatus);
+        removeEligibilityStatusIfAppropriate(eligibilityStatus, builder);
+        return builder
+                .qualifyingBenefitEligibilityStatus(qualifyingBenefitEligibilityStatus)
+                .dateOfBirthOfChildren(childrenDobs)
+                .build();
+    }
+
+    private static void removeEligibilityStatusIfAppropriate(EligibilityStatus eligibilityStatus,
```
dwybourn: This could go into aValidDecisionBuilder, unless that break tests which use it directly?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
mikeSmith08: That case is covered and tested. In that scenario we would fall through this method and the claim would not be expired, it would then either go to `handleLossOfQualifyingBenefitStatus` if they were not eligible with DWP or `handleNoLongerEligibleForSchemeAsNoChildrenAndNotPregnant` if they were still eligible.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: I have double checked this with Karen and this is correct for the scenario where the claimant was pregnant but is now no longer pregnant (but would have been in the new cycle). There is another story (HTBHF-1503) to do the scenario where they haven't told DWP about their children yet and they are 8 (or 12 weeks can't remember which) after their due date.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -94,17 +109,75 @@ void shouldSuccessfullyProcessMessageAndTriggerPaymentWhenClaimantIsEligible() {
         verify(paymentCycleService).updatePaymentCycle(context.getCurrentPaymentCycle(), decision);
         MessagePayload expectedPayload = MessagePayloadFactory.buildMakePaymentMessagePayload(context.getCurrentPaymentCycle());
         verify(messageQueueClient).sendMessage(expectedPayload, MessageType.MAKE_PAYMENT);
-        verifyZeroInteractions(claimRepository, pregnancyEntitlementCalculator);
+        verifyZeroInteractions(claimRepository, pregnancyEntitlementCalculator, childDateOfBirthCalculator, childDateOfBirthCalculator);
+    }
+
+    @ParameterizedTest(name = "Qualifying benefit status={0}")
+    @ValueSource(strings = {"CONFIRMED", "NOT_CONFIRMED"})
+    void shouldExpireClaimAndSendNoLongerOnSchemeEmailWhenEligibleWithNoChildrenAndNotPregnant(
```
YetAnotherMatt: I don't understand how someone could be eligible when their `QualifyingBenefitEligibilityStatus` is `NOT_CONFIRMED`? There's also a comment to the same effect on line 138

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -94,17 +109,75 @@ void shouldSuccessfullyProcessMessageAndTriggerPaymentWhenClaimantIsEligible() {
...
     }
 
     @Test
     void shouldExpireClaimAndSendNoLongerOnSchemeEmailWhenEligibleWithNoChildrenAndNotPregnant() {
         //Given
-        DetermineEntitlementMessageContext context = buildMessageContext(NOT_PREGNANT);
+        //Test for HTBHF-2185 has the following context:
+        // Previous cycle: children exist and are under 4, not pregnant
+        // Current cycle: no children and not pregnant
+        LocalDate previousCycleExpectedDeliveryDate = NOT_PREGNANT;
```
YetAnotherMatt: `previousCycleExpectedDeliveryDate` and `currentCycleExpectedDeliveryDate` could be inlined.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -94,17 +109,75 @@ void shouldSuccessfullyProcessMessageAndTriggerPaymentWhenClaimantIsEligible() {
...
+        LocalDate currentCycleExpectedDeliveryDate = NOT_PREGNANT;
+        List<LocalDate> previousCycleChildrenDobs = SINGLE_THREE_YEAR_OLD;
+        List<LocalDate> currentCycleChildrenDobs = NO_CHILDREN;
+        DetermineEntitlementMessageContext context = buildMessageContext(
+                previousCycleExpectedDeliveryDate,
+                currentCycleExpectedDeliveryDate,
+                previousCycleChildrenDobs,
+                currentCycleChildrenDobs);
```
YetAnotherMatt: Minor, but I was wondering whether
```
        DetermineEntitlementMessageContext context = buildMessageContext(
                previousCycleExpectedDeliveryDate,
                currentCycleExpectedDeliveryDate,
                previousCycleChildrenDobs,
                currentCycleChildrenDobs);
```
might be easier to read as
```
        DetermineEntitlementMessageContext context = buildMessageContext(
                previousCycleExpectedDeliveryDate,
                previousCycleChildrenDobs,
                currentCycleExpectedDeliveryDate,
                currentCycleChildrenDobs);
```
?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: They could be inlined however I found it much clearer to have the  variables set out like this as I always had to figure out which was which when the `buildMessageContext` was called below.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: I think both are pretty much the same, I quite like looking at the two EDDs and two lists of DOBs next to each other as those were the main things you were comparing to each other.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -146,16 +224,162 @@ void shouldUpdateClaimWhenClaimantIsNotEligible(EligibilityStatus eligibilitySta
...
+        //Then
+        runCommonVerifications(context, decision, message, messageStatus, ClaimStatus.PENDING_EXPIRY);
+        verify(determineEntitlementNotificationHandler).sendClaimNoLongerEligibleEmail(context.getClaim());
+        verify(pregnancyEntitlementCalculator).isEntitledToVoucher(EXPECTED_DELIVERY_DATE_IN_TWO_MONTHS, context.getCurrentPaymentCycle().getCycleStartDate());
+        verifyZeroInteractions(childDateOfBirthCalculator);
+    }
+
+    //HTBHF-1757 Claimant had children in previous cycle who would be under 4 in this cycle but are not returned from DWP
+    @Test
+    void shouldUpdateClaimToPendingExpiryAndSendNoLongerEligibleEmailWhenClaimantIsNotEligibleWithNoChildrenInCurrentCycle() {
```
dwybourn: Could we have a display name for this test and the one above. Otherwise it's quiet difficult to see where they differ, especially as they have the same name.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -146,16 +224,162 @@ void shouldUpdateClaimWhenClaimantIsNotEligible(EligibilityStatus eligibilitySta
...
+        LocalDate currentCycleExpectedDeliveryDate = NOT_PREGNANT;
+        DetermineEntitlementMessageContext context = buildMessageContext(
+                previousCycleExpectedDeliveryDate,
+                currentCycleExpectedDeliveryDate,
+                previousCycleChildrenDobs,
+                currentCycleChildrenDobs);
+        given(messageContextLoader.loadDetermineEntitlementContext(any())).willReturn(context);
+        LocalDate currentPaymentCycleStartDate = context.getCurrentPaymentCycle().getCycleStartDate();
+        LocalDate previousPaymentCycleStartDate = context.getPreviousPaymentCycle().getCycleStartDate();
+        lenient().when(pregnancyEntitlementCalculator.isEntitledToVoucher(NOT_PREGNANT, currentPaymentCycleStartDate)).thenReturn(false);
```
dwybourn: Why is lenient needed here?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: Done.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: Done.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: Because this an the next line are calling the same method on the same method twice, if I remove lenient then Mockito instantly fails the test.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
mikeSmith08: Sorted now.

PR #359, opened 2019-10-21
Feature/call ga when theres a new claim

dwybourn:
- MIReporter now calls google analytics client to report the new claim.
- Added an integration tests that puts a REPORT_CLAIM message on the queue and invokes the scheduler.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -93,7 +85,7 @@ void shouldCreateReportPropertiesForNewClaim() {
                 entry("cd7", postcodeData.getOutcode()),
                 entry("cd8", postcodeData.getParliamentaryConstituency()),
                 entry("cd9", postcodeData.getCcg()), // Clinical Commissioning Group
-                entry("cd10", postcodeData.getCodes().getCcg()), // Clinical Commissioning Group code
+                entry("cd10", postcodeData.getCodes().getCcg()), // Clinical Commissioning Group code`
```
YetAnotherMatt: Slightly spurious addition of ` to the comment :)

dwybourn: Not sure how that happened, removed now.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -56,6 +56,12 @@
     public static final LocalDate NEARLY_FOUR_YEAR_OLD = LocalDate.now().minusYears(4).plusWeeks(2);
     public static final List<LocalDate> TWO_CHILDREN = List.of(SIX_MONTH_OLD, THREE_YEAR_OLD);
     public static final List<LocalDate> NO_CHILDREN = emptyList();
+    public static final List<LocalDate> TWO_CHILDREN_UNDER_ONE = List.of(LocalDate.now().minusDays(1), LocalDate.now().minusMonths(11));
+    public static final List<LocalDate> TWO_CHILDREN_BETWEEN_ONE_AND_FOUR = List.of(LocalDate.now().minusYears(2), LocalDate.now().minusYears(3));
+    public static final List<LocalDate> ONE_CHILD_FOUR_YEARS_OLD = List.of(LocalDate.now().minusYears(4));
+    public static final List<LocalDate> ONE_CHILD_UNDER_ONE_AND_ONE_CHILD_BETWEEN_ONE_AND_FOUR
```
mikeSmith08: I think we should try to use some of the existing constants here so this doesn't explode. For example, you should be able to use `TWO_CHILDREN` instead of creating `ONE_CHILD_UNDER_ONE_AND_ONE_CHILD_BETWEEN_ONE_AND_FOUR`, or if you prefer the new name, then get rid of the original one, but there is no need for both. Can you double check the others too please?

dwybourn: I'ved removed the TWO_CHILDREN constant, the rest are different enough to leave.

PR #363, opened 2019-10-22
HTBHF-2040 Added message, context and loader for REPORT_PAYMENT.

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -97,4 +97,14 @@ public void processPaymentMessages() {
     public void processReportClaimMessages() {
         messageProcessor.processMessagesOfType(MessageType.REPORT_CLAIM);
     }
+
+    @Scheduled(cron = EVERY_30_SECONDS)
```
YetAnotherMatt: Shall we turn the offset-by-5 schedule into a constant and use that for all MI reports?

mikeSmith08:
Looks like your commit message and title of this PR are incorrect as you're talking about `REPORT_PAYLOAD` rather than `REPORT_PAYMENT`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
```diff
@@ -127,6 +127,20 @@ public ReportClaimMessageContext loadReportClaimMessageContext(Message message)
                 .build();
     }
 
+    public ReportPaymentMessageContext loadReportPaymentMessageContext(Message message) {
+        ReportPaymentMessagePayload payload = payloadMapper.getPayload(message, ReportPaymentMessagePayload.class);
+        Claim claim = getAndCheckClaim(payload.getClaimId());
+        Optional<PaymentCycle> paymentCycle = payload.getPaymentCycleId().flatMap(id -> paymentCycleRepository.findById(id));
```
mikeSmith08: Why does this need to complexity of `flatMap` when you could just do `paymentCycleRepository.findById(payload.getPaymentCycleId())` and no `flatMap` required?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/ReportPaymentMessageContext.java
```diff
@@ -0,0 +1,22 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+
+@Value
+@Builder
+public class ReportPaymentMessageContext {
+    private Claim claim;
+    private Optional<PaymentCycle> paymentCycle;
```
mikeSmith08: Why is the PaymentCycle Optional?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/PaymentAction.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.dhsc.htbhf.claimant.reporting;
+
+/**
+ * Action that triggered a payment event.
+ */
+public enum PaymentAction {
```
mikeSmith08: Might be good to have a description of what the enum values mean because I have no idea.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
mikeSmith08: What is the "offset-by-5 schedule"?

YetAnotherMatt: This (for "MessageProcessor:REPORT_CLAIM"):
```
    @Scheduled(cron = "05/35 * * * * *") // every 30 seconds, offset from other schedules by 5 seconds to prevent clashes with CREATE_NEW_CARD for new claims
```

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/ReportPaymentMessageContext.java
YetAnotherMatt: Because for additional pregnancy payments we _might_ not have a PaymentCycle

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -97,4 +97,14 @@ public void processPaymentMessages() {
...
+    @SchedulerLock(
+            name = "Process REPORT_PAYMENT messages",
+            lockAtLeastForString = FIFTEEN_SECONDS,
+            lockAtMostForString = ONE_HOUR)
```
mikeSmith08: Maybe not for this PR, but it would probably be a good idea to have our own annotation which defaults the values for `lockAtLeastForString` and `lockAtMostForString` to these values because the majority of the time we want them all to be the same. If we did this, it would also mean that any that validly need to differ stick out clearly from the rest.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
mikeSmith08: Ah yes, let's have a constant for that.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
YetAnotherMatt: Because `payload.getPaymentCycleId()` returns an optional. The alternative would be something like
```
Optional<PaymentCycle> paymentCycle = payload.getPaymentCycleId().ifPresentOrElse(id -> paymentCycleRepository.findById(id), Optional.empty())
```

dwybourn:
> Looks like your commit message and title of this PR are incorrect as you're talking about `REPORT_PAYLOAD` rather than `REPORT_PAYMENT`.

Fixed the title, will fix the commit message when I squash and merge the PR.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/PaymentAction.java
dwybourn: Added a comment for each value.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
dwybourn: Will do in it's own PR, potentially with a custom annotation to group the Schedule annotations together.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
mikeSmith08: Yep, think I prefer `ifPresentOrElse`.

dwybourn: As discussed, sticking with flatMap.

PR #368, opened 2019-10-24
Feature/create report payment properties

dwybourn:
- Made payment cycle no longer optional as it will always be provided.
- Added payment amounts for each category. The creator of the REPORT_PAYMENT message will determine if this comes from the payment cycle or somewhere else (as is the case for additional pregnancy voucher).
-  Creating payment report properties.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -205,6 +242,22 @@ private ReportClaimMessageContext aReportClaimMessageContext(LocalDateTime times
                 .build();
     }
 
+    private ReportPaymentMessageContext aReportPaymentMessageContext(LocalDateTime timestamp,
```
mikeSmith08: Test data factory?

dwybourn: Wouldn't make sense here as every value is being explicitly set to a value that we're testing against.

PR #371, opened 2019-10-25
HTBHF-2012 Handling null children when report events. Moved private n…

dwybourn:
- Handling null children when report events 
- Moved private number of children methods into ChildDateOfBirthCalculator


api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactoryTest.java
```diff
@@ -119,19 +119,6 @@ void shouldImposeMaximumValueForQueueTime() {
         assertThat(queueTime).isLessThanOrEqualTo(maxQueueTime);
     }
 
-    @Test
-    void shouldIncludeMetricsForZeroChildren() {
```
YetAnotherMatt: Isn't this test still valid?

PR #374, opened 2019-10-25
Feature/refactor voucher entitlement

dwybourn:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/PaymentCycleVoucherEntitlement.java
```diff
@@ -22,6 +22,9 @@
     private int vouchersForPregnancy;
     private int totalVoucherEntitlement;
     private int singleVoucherValueInPence;
+    private int totalChildrenUnderOneVoucherValueInPence;
```
YetAnotherMatt: We need to bear in mind that we already have PaymentCycles in the database. I think we should either update the entities in the database, or have getter methods that perform the calculation (and not store this duplicate information). Or, we just live with performing the calculation in two places.

dwybourn: I'm personally against editing production unless absolutely necessary. I would be in favour of just rejecting the PR.

PR #376, opened 2019-10-28
HTBHF-2263 MessageRepository now excludes messages dated in the future

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/MessageRepository.java
```diff
@@ -17,8 +18,16 @@
...
-            + "where m.messageType = :messageType "
+            + "where m.messageType = :messageType and m.messageTimestamp < :olderThan "
             + "order by m.messageTimestamp")
-    List<Message> findAllMessagesByTypeOrderedByDate(@Param("messageType") MessageType messageType, Pageable pageable);
+    List<Message> findAllMessagesByTypeOlderThan(
+            @Param("messageType") MessageType messageType,
+            @Param("olderThan") LocalDateTime olderThan,
+            Pageable pageable);
+
+    default List<Message> findAllMessagesDueForProcessingByType(
```
mikeSmith08: I'm not too keen on the phrase `dueForProcessing`, its the kind of term that means something if you know what it means (if that makes sense). How about something like `findAllMessagesOlderThanNow` or `findAllPastMessages` as you can understand exactly what both of those mean, although they might sound a bit clunky, but they are understandable.

mikeSmith08:
Are we planning to have future dated messages then, is this the point of this PR?

YetAnotherMatt:
> Are we planning to have future dated messages then, is this the point of this PR?

Hi @mikeSmith08 . Yes, the next PR provides a delay for retrying failed messages - using an exponential back-off strategy. There are upcoming stories about delaying messages when, for instance, we want to delay sending an email about upcoming changes for a day or two.

PR #377, opened 2019-10-28
HTBHF-2016 Reporting change in claim status to google analytics.

dwybourn:
A change in a claim's status is now reported to google analytics.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -85,9 +93,11 @@ private void handleDecision(Claim claim, PaymentCycle previousPaymentCycle, Paym
             createMakePaymentMessage(currentPaymentCycle);
         } else if (claim.getClaimStatus() == ACTIVE) {
             if (shouldExpireClaim(decision, previousPaymentCycle, currentPaymentCycle)) {
-                expireClaim(claim);
+                updateClaimStatus(claim, ClaimStatus.EXPIRED);
+                eventAuditor.auditExpiredClaim(claim);
+                claimMessageSender.sendReportClaimMessage(claim, emptyList(), UPDATED_FROM_ACTIVE_TO_EXPIRED);
```
YetAnotherMatt: shouldn't we pass the list of children's dob here? I know that we shouldn't have any children at this point, but just in case that changes...

dwybourn: Updated to include the children from the decision.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -123,19 +130,21 @@ private void createMakePaymentMessage(PaymentCycle paymentCycle) {
...
+    private void handleNoLongerEligibleForSchemeAsNoChildrenAndNotPregnant(Claim claim, List<LocalDate> dateOfBirthOfChildren) {
+        expireClaim(claim, dateOfBirthOfChildren);
         determineEntitlementNotificationHandler.sendNoChildrenOnFeedClaimNoLongerEligibleEmail(claim);
     }
 
-    private void expireClaim(Claim claim) {
+    private void expireClaim(Claim claim, List<LocalDate> dateOfBirthOfChildren) {
         updateClaimStatus(claim, ClaimStatus.EXPIRED);
         eventAuditor.auditExpiredClaim(claim);
+        claimMessageSender.sendReportClaimMessage(claim, dateOfBirthOfChildren, UPDATED_FROM_ACTIVE_TO_EXPIRED);
```
YetAnotherMatt: can we be sure they are active currently? I think it's possible that they were pending expiry. (I think this would mean that they became eligible again this cycle, but it happened to coincide with their children turning 4)

dwybourn: The method is currently only called if the claim is active, I'll rename the private method to expireActiveClaim which should prevent it being called when we handle the case of pending_expiry to expired.

dwybourn: Went for renaming the private method to expireActiveClaim

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/ClaimAction.java
```diff
@@ -5,5 +5,9 @@
  */
 public enum ClaimAction {
     NEW,
-    REJECTED
+    REJECTED,
+    UPDATED_FROM_NEW_TO_ACTIVE,
+    UPDATED_FROM_ACTIVE_TO_PENDING_EXPIRY,
+    UPDATED_FROM_ACTIVE_TO_EXPIRED,
+    UPDATED_FROM_PENDING_EXPIRY_TO_EXPIRED
```
mikeSmith08: We should have `UPDATED_FROM_PENDING_EXPIRY_TO_ACTIVE` here too.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/NewCardService.java
```diff
@@ -33,16 +38,18 @@
      * </p>
      *
      * @param claim the claim to create a new card for.
+     * @param datesOfBirthOfChildren dates of birth of the claimant's children
      */
     @SuppressWarnings("PMD.NullAssignment")
     @Transactional
-    public void createNewCard(Claim claim) {
+    public void createNewCard(Claim claim, List<LocalDate> datesOfBirthOfChildren) {
```
mikeSmith08: It feels like passing the dates of birth through all these message calls is quite messy when we could simply load the dates of birth from the PaymentCycle when the MI reporting message is loaded to be processed. @YetAnotherMatt - this has sparked a bit of debate with Dave and I and we can see pros and cons of both, which way do you prefer?

YetAnotherMatt: I don't think we can do that for a new card, sadly - the payment cycle may not exist yet. If we start storing the full eligibility and entitlement decision on the claim then we could use that.

YetAnotherMatt: I might be wrong - not able to spend much time on it right now I'm afraid. Will look again tomorrow.

dwybourn: I'm sceptical of getting the information later as it creates an order dependency which might forget if we make changes at a later date.
Here we can be 100% sure that we have the necessary information it's being passed in directly.

YetAnotherMatt: Once we have the decision from DWP then we can be sure of having the children. Could we add a TODO statement to use that once it's available? (HTBHF-2457)

dwybourn: Added Todo statement.

PR #383, opened 2019-10-30
Feature/load updated claimant fields

dwybourn:
Updated ReportClaimMessagePayload and MessageContextLoader to handle UpdatableClaimantFields


mikeSmith08:
What's the point of this change, I can't see what its used for?

PR #384, opened 2019-10-30
NO-JIRA Added the ability to fast-forward the age of entities in a db, not just in integration tests

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -22,12 +22,12 @@
 @Component
 public class MessageProcessorScheduler {
 
-    private static final String EVERY_30_SECONDS = "*/2 * * * * *";
+    private static final String EVERY_30_SECONDS = "*/30 * * * * *";
```
YetAnotherMatt: We should probably go back to using properties for this to make sure we don't inadvertently change it again.

PR #390, opened 2019-11-01
HTBHF-1296 Added method to get claims with cards to be cancelled.

dwybourn:
Added method to the ClaimRepository which gets all the the claims with a card status of PENDING_CANCELLATION and a card status timestamp before a given period.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepository.java
```diff
@@ -42,6 +48,20 @@
     @Query("SELECT claim.id FROM Claim claim where claim.claimStatus = 'NEW'")
     List<UUID> getNewClaimIds();
 
+    /**
+     * Gets the claims that have a card_status of PENDING_CANCELLATION and a card_status_timestamp older than or equal to than the given period.
+     *
+     * @param period length of time the card status has been PENDING_CANCELLATION.
+     * @return list of claims
+     */
+    default List<Claim> getClaimsWithCardStatusPendingCancellationForMoreThan(Period period) {
```
mikeSmith08: Possibly `...OlderThan`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepository.java
```diff
@@ -42,6 +48,20 @@
...
+        LocalDateTime timestamp = LocalDateTime.now().minus(period);
+        return getClaimsWithCardStatusSinceBeforeGivenTimestamp(PENDING_CANCELLATION, timestamp);
+    }
+
+    @Query("SELECT claim FROM Claim claim where claim.cardStatus = :cardStatus and claim.cardStatusTimestamp <= :date")
+    List<Claim> getClaimsWithCardStatusSinceBeforeGivenTimestamp(@Param("cardStatus") CardStatus cardStatus, @Param("date") LocalDateTime localDateTime);
```
mikeSmith08: Don't think you need the word `Since` in this method name, `getClaimsWithCardStatusBeforeGivenTimestamp` works pretty well.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepositoryTest.java
```diff
@@ -321,4 +321,25 @@ void shouldThrowExceptionWhenClaimDoesNotExist() {
...
+        Claim pendingClaimLessThan16Weeks = aClaimWithCardStatusAndCardStatusTimestamp(CardStatus.PENDING_CANCELLATION, now.minusWeeks(15));
+        claimRepository.saveAll(List.of(activeClaim, pendingClaimOlderThan16Weeks, pendingClaimLessThan16Weeks));
+
+        List<Claim> claims = claimRepository.getClaimsWithCardStatusPendingCancellationForMoreThan(Period.ofWeeks(numberOfWeeks));
+
+        assertThat(claims).containsOnly(pendingClaimOlderThan16Weeks);
+    }
+
+    @Test
+    void shouldReturnEmptyListWhenThereAreNoClaimsThatHaveBeenInPendingCancellationForMoreThanGivenPeriod() {
```
mikeSmith08: This test would be better if you added some claims to the database, or call it `...NoClaimsInDatabase...`, but I guess that is then a different test so maybe have another test to cover the two scenarios.

dwybourn: Added in one claim that's not pending cancellation.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepository.java
dwybourn: Removed since.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/repository/ClaimRepository.java
dwybourn: Changed to olderThan

PR #391, opened 2019-11-01
HTBHF-1296 Building email message payload for card to be cancelled em…

dwybourn:
Building email message payload for card to be cancelled emails.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
@@ -3,6 +3,7 @@ package uk.gov.dhsc.htbhf.claimant.communications;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleVoucherEntitlement;
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
 import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
```diff
@@ -57,6 +58,18 @@ public EmailMessagePayload buildEmailMessagePayload(PaymentCycle paymentCycle, E
                 .build();
     }
 
+    public EmailMessagePayload buildEmailMessagePayloadWithFirstAndLastNameOnly(Claim claim, EmailType emailType) {
```
mikeSmith08: This might be able to be refactored with the code in `DetermineEntitlementNotificationHandler`. Similarly for the unit test below with `DetermineEntitlementNotificationHandlerTest`.

dwybourn: Moved the private method out of DetermineEntitlementNotificationHandler.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/EmailMessagePayloadFactory.java
```diff
@@ -32,8 +33,20 @@ public static String formatPaymentAmountSummary(String summaryTemplate, int numb
         return formatPaymentAmount(summaryTemplate, numberOfVouchers, voucherAmountInPence);
     }
 
-    public static String formatRegularPaymentAmount(int numberOfVouchers, int voucherAmountInPence) {
```
dwybourn: This one line method was only called in one place so it's been inlined to reduce the number of methods on the class.

PR #393, opened 2019-11-04
HTBHF-1296 Setting card status to PENDING_CANCELLATION for non eligib…

dwybourn:
Setting card status to PENDING_CANCELLATION for non eligible claims.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
@@ -24,6 +24,7 @@ import java.time.LocalDate;
 import java.util.List;
 import javax.transaction.Transactional;
 
+import static uk.gov.dhsc.htbhf.claimant.entity.CardStatus.PENDING_CANCELLATION;
 import static uk.gov.dhsc.htbhf.claimant.message.MessageStatus.COMPLETED;
 import static uk.gov.dhsc.htbhf.claimant.message.MessageType.DETERMINE_ENTITLEMENT;
 import static uk.gov.dhsc.htbhf.claimant.model.ClaimStatus.ACTIVE;
@@ -98,6 +99,8 @@ public class DetermineEntitlementMessageProcessor implements MessageTypeProcesso
             } else {
                 handleNoLongerEligibleForSchemeAsNoChildrenAndNotPregnant(claim, decision.getDateOfBirthOfChildren());
             }
+
+            setCardStatusToPendingCancellation(claim);
         }
         //TODO HTBHF-1296: If not ACTIVE, PENDING_EXPIRY will be moved to EXPIRED after 16 weeks.
     }
@@ -145,6 +148,11 @@ public class DetermineEntitlementMessageProcessor implements MessageTypeProcesso
         updateClaimStatus(claim, ClaimStatus.EXPIRED);
         eventAuditor.auditExpiredClaim(claim);
         claimMessageSender.sendReportClaimMessage(claim, dateOfBirthOfChildren, UPDATED_FROM_ACTIVE_TO_EXPIRED);
+    }
+
+    private void setCardStatusToPendingCancellation(Claim claim) {
+        claim.updateCardStatus(PENDING_CANCELLATION);
+        claimRepository.save(claim);
     }
 
     private void updateClaimStatus(Claim claim, ClaimStatus claimStatus) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -98,6 +99,8 @@ private void handleDecision(Claim claim, PaymentCycle previousPaymentCycle, Paym
             } else {
                 handleNoLongerEligibleForSchemeAsNoChildrenAndNotPregnant(claim, decision.getDateOfBirthOfChildren());
             }
+
+            setCardStatusToPendingCancellation(claim);
```
YetAnotherMatt: I don't think this is in the right place. This seems to me to always set active claims to pending_expiry?

YetAnotherMatt: I take it back. This is in the right place.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -123,7 +125,7 @@ void shouldSuccessfullyProcessMessageAndTriggerPaymentWhenClaimantIsEligible() {
 
     @ParameterizedTest(name = "Qualifying benefit status={0}")
     @ValueSource(strings = {"CONFIRMED", "NOT_CONFIRMED"})
-    void shouldExpireClaimAndSendNoLongerOnSchemeEmailWhenIneligibleWithNoChildrenAndNotPregnant(
+    void shouldExpireClaimAndSendNoLongerOnSchemeEmailAndSetCardStatusToPendingCancellationWhenIneligibleWithNoChildrenAndNotPregnant(
```
YetAnotherMatt: This name is getting to be too long now. I think we should go with `shouldExpireClaimWhenIneligibleWithNoChildrenAndNotPregnant` and either update the name in the `@ParameterizedTest` annotation or add a comment. We can't keep adding every change to the method name.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
```diff
@@ -165,7 +167,8 @@ void shouldExpireClaimAndSendNoLongerOnSchemeEmailWhenIneligibleWithNoChildrenAn
 
     @ParameterizedTest(name = "Expected delivery date previous cycle={0}")
     @MethodSource("provideArgumentsForPreviousCycleExpectedDeliveryDate")
-    void shouldExpireClaimAndSendNoLongerOnSchemeEmailWhenChildDisappearsFromFeedAndNotPregnant(LocalDate previousCycleExpectedDeliveryDate) {
+    void shouldExpireClaimAndSendNoLongerOnSchemeEmailAndSetCardStatusToPendingCancellationWhenChildDisappearsFromFeedAndNotPregnant(
```
YetAnotherMatt: Same with all the method names here...

api/src/test/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessorTest.java
dwybourn: Renamed to be shorter. Didn't add a comment seeing as the tests themselves are descriptive of the expected behaviour.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -147,6 +150,10 @@ private void expireActiveClaim(Claim claim, List<LocalDate> dateOfBirthOfChildre
         claimMessageSender.sendReportClaimMessage(claim, dateOfBirthOfChildren, UPDATED_FROM_ACTIVE_TO_EXPIRED);
     }
 
+    private void setCardStatusToPendingCancellation(Claim claim) {
+        claim.updateCardStatus(PENDING_CANCELLATION);
```
mikeSmith08: You should call `claimRepository.save(claim)` here.

dwybourn: Done.

YetAnotherMatt: Should we be looking to save the claim once rather than in two separate places?

mikeSmith08: Possibly, but that might be a bit more tricky than is practicable as not all the routes through this code call save.

PR #394, opened 2019-11-04
NO-JIRA upgraded spring boot to 2.2.0, added an ObjectMapper bean

YetAnotherMatt:
Just upgrading Spring without adding an ObjectMapper caused the serialisation of dates to go wrong. I'm happier explicitly defining how they should be serialised anyway. The new version of Jackson also changed the exceptions thrown by ObjectMapper.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -17,7 +17,7 @@ buildscript {
 }
 
 plugins {
-    id "org.springframework.boot" version "2.1.9.RELEASE"
+    id "org.springframework.boot" version "2.2.0.RELEASE"
     id 'io.franzbecker.gradle-lombok' version '1.14'
     id "org.owasp.dependencycheck" version "4.0.0.1"
     id "com.github.spotbugs" version "2.0.0"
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceApplication.java
```diff
@@ -47,6 +57,20 @@ public Clock clock() {
         return Clock.systemDefaultZone();
     }
 
+    @Bean
+    public ObjectMapper objectMapper() {
```
mikeSmith08: I'm personally against having to define these things (convention over configuration and all that), but if we must do this, could we add this to the java-common package instead so that we don't have to do this kind of boiler plate code in all our projects. I'd be interested to know what failed and why these date formats are not standard anymore too.

YetAnotherMatt: We don't have to - this was my mistake.

YetAnotherMatt:
The build should start passing again once the java-common pr has been merged.

PR #398, opened 2019-11-08
HTBHF-2390 Renamed messageTimestamp property in entity to processAfte…

dwybourn:
Renamed messageTimestamp property in Message entity to processAfterTimestmap.
Added process_after_timestamp column which copies the data from message_timestamp.

This is part one of the rename process which allows no downtime during green/blue deployment. 
The next PR will remove the message_timestamp column.


db/src/main/resources/db/migration/V1_043__create_process_after_timestamp_column.sql
```diff
@@ -0,0 +1,4 @@
+alter table message_queue add column process_after_timestamp timestamp;
+
+-- noinspection SqlWithoutWhere
```
dwybourn: Intellij will warn about update statements that don't have a where clause because it will update every row. However, that is exactly what we want in this situation so suppressing this warning.

YetAnotherMatt: I'd suggest including a not null check on process_after instead

dwybourn: Done.

PR #399, opened 2019-11-08
HTBHF2564 converted message sending delay from Period to Duration

YetAnotherMatt:


mikeSmith08:
Can I ask why the change?

YetAnotherMatt:
> Can I ask why the change?

Because a Period has to be longer than a day.

PR #400, opened 2019-11-08
Feature/remove message timestamp

dwybourn:


db/src/main/resources/db/migration/V1_044__remove_message_timestamp.sql
```diff
@@ -0,0 +1 @@
+alter table message_queue drop column message_timestamp;
```
YetAnotherMatt: To be safe, we also need to re-run the update of process_after

dwybourn: Updated.

PR #407, opened 2019-11-13
HTBHF-2377 Refactored DetermineEntitlementMessageProcessor ineligible…

dwybourn:
DetermineEntitlementMessageProcessor currently has 10 dependencies and was about to have one more with some upcoming work. It was therefore decided to split out some of the logic into a separate class. Doing this has halved the number of dependencies and allowed us to remove the suppression of PMD's TooManyMethods warning.

Ineligible decisions are now handled in the new IneligibleEntitlementDecisionHandler.

The code and tests have been copied directly from one class to another, no logic has been changed.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/IneligibleEntitlementDecisionHandler.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/IneligibleEntitlementDecisionHandler.java
@@ -1,0 +1,105 @@
+package uk.gov.dhsc.htbhf.claimant.entitlement;
+
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Component;
+import uk.gov.dhsc.htbhf.claimant.communications.DetermineEntitlementNotificationHandler;
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entitlement/IneligibleEntitlementDecisionHandler.java
```diff
@@ -0,0 +1,105 @@
...
+import java.time.LocalDate;
+import java.util.List;
+
+import static uk.gov.dhsc.htbhf.claimant.entity.CardStatus.PENDING_CANCELLATION;
+import static uk.gov.dhsc.htbhf.claimant.reporting.ClaimAction.UPDATED_FROM_ACTIVE_TO_EXPIRED;
+import static uk.gov.dhsc.htbhf.claimant.reporting.ClaimAction.UPDATED_FROM_ACTIVE_TO_PENDING_EXPIRY;
+
+@Component
+@AllArgsConstructor
+public class IneligibleEntitlementDecisionHandler {
```
YetAnotherMatt: Suggest calling this something like `EligibilityDecisionHandler`

dwybourn: Renamed to EligibilityDecisionHandler and moved to new eligibility package.

PR #409, opened 2019-11-14
HTBHF-2377 Created method for sending report a birth email with a delay.

dwybourn:
The next PR updated DetermineEntitlementProcessor to make use of this new method.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimMessageSenderTest.java
```diff
@@ -55,8 +56,8 @@
     private ClaimMessageSender claimMessageSender;
 
     @BeforeEach
-    private void setup() {
```
dwybourn: Made package-private purely to remove intellij warning.

api/src/main/resources/application.yml
```diff
@@ -94,6 +94,7 @@ card-cancellation:
     minimum-lock-time: "PT1M"
     maximum-lock-time: "PT59M"
 
+report-a-birth-message-delay-duration: P0D
```
YetAnotherMatt: Should this belong within 'payment-cycle'? Also, to be in keeping with 'change-in-payment-email-delay' we should remove '-duration' from the name.

dwybourn: Moved under payment-cycle and removed the -duration part of the name.

PR #410, opened 2019-11-14
HTBHF-2572 Add identity_and_eligibility_response to the payment cycle table and entity.

mikeSmith08:
- Added default values for the three existing values of QualifyingBenefitEligibilityStatus so it can be replaced in the next PR.
 - Also added matching childrens' dates of birth to the test data factory.


db/src/main/resources/db/migration/V1_045__create_identity_and_eligibility_outcome_column.sql
```diff
@@ -0,0 +1,46 @@
...
+    "eligibilityStatus": "confirmed",
+    "deathVerificationFlag": "n/a",
+    "mobilePhoneMatch": "matched",
+    "emailAddressMatch": "matched",
+    "addressLine1Match": "matched",
+    "postcodeMatch": "matched",
+    "pregnantChildDOBMatch": "not_supplied",
+    "qualifyingBenefits": "universal_credit",
+    "householdIdentifier": "",
+    "dobOfChildrenUnder4": [],
```
YetAnotherMatt: Strictly speaking we should probably use the list of children's dob from the claim, but as this is only used for reference I don't think we need to worry.

PR #423, opened 2019-11-19
HTBHF-1413 Renamed CREATE_NEW_CARD message type REQUEST_NEW_CARD and …

dwybourn:
- Renamed CREATE_NEW_CARD message type to REQUEST_NEW_CARD.
- Renamed new card related services and pojos to follow the pattern of RequestNewCard*

This has been done in preparation for the creation of a new message type and processor which will be handling all of the new card steps that occur after requesting the card, e.g. saving to claim, creating payment cycle, auditing, etc.
Once this new message and processor exists, the REQUEST_NEW_CARD message processor will only be requesting a new card, hence the rename.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
```diff
@@ -1,7 +1,7 @@
 package uk.gov.dhsc.htbhf.claimant.message;
 
 public enum MessageType {
-    CREATE_NEW_CARD,
+    REQUEST_NEW_CARD,
```
YetAnotherMatt: Strictly speaking we should have some SQL to migrate CREATE_NEW_CARD messages to REQUEST_NEW_CARD. However we don't have any live claims so we can probably get away without it.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/MessageContextLoader.java
```diff
@@ -73,11 +73,11 @@ public MakePaymentMessageContext loadMakePaymentContext(Message message) {
      * @param message The message to inflate.
      * @return A wrapper object with the inflated objects
      */
-    public NewCardMessageContext loadNewCardContext(Message message) {
-        NewCardRequestMessagePayload payload = payloadMapper.getPayload(message, NewCardRequestMessagePayload.class);
+    public RequestNewCardMessageContext loadNewCardContext(Message message) {
```
mikeSmith08: This method should probably be renamed to `loadRequestNewCardContext` otherwise it will clash when the new message is created which will need a similar method like `loadCreateNewCardContext`.

dwybourn: Renamed.

PR #426, opened 2019-11-20
HTBHF-2570 Create EligibilityAndEntitlementDecisionFactory for building…

mikeSmith08:
…EligibilityAndEntitlementDecisions.

This is a slightly modified version of what currently exists in EligibilityAndEntitlementServiceV1 which has been modified so it can be used by a v2 version of the service. Only a small modification needs to be made to the v1 service to be able to use this, so this will be the next PR. Hopefully by breaking out all these parts of the v1 service so they can be used in the v2 service, when it comes to implementing the v2 service it will be quite small and simple.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementDecisionFactoryTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleVoucherEntitlementTestDataFactory.aPaymentCycleVoucherEntitlementWithVouchers;
+import static uk.gov.dhsc.htbhf.claimant.testsupport.PaymentCycleVoucherEntitlementTestDataFactory.aPaymentCycleVoucherEntitlementWithZeroVouchers;
+import static uk.gov.dhsc.htbhf.dwp.testhelper.TestConstants.NO_HOUSEHOLD_IDENTIFIER_PROVIDED;
+import static uk.gov.dhsc.htbhf.dwp.testhelper.v2.IdentityAndEligibilityResponseTestDataFactory.anIdentityMatchedEligibilityNotConfirmedResponse;
+
+class EligibilityAndEntitlementDecisionFactoryTest {
+
+    private EligibilityAndEntitlementDecisionFactory factory = new EligibilityAndEntitlementDecisionFactory();
+    private IdentityAndEligibilityResponse eligibilityResponse =
+            ClaimantServiceIdentityAndEligibilityResponseTestDataFactory.anAllMatchedEligibilityConfirmedUCResponseWithHouseholdIdentifier();
```
mikeSmith08: Cannot use a static import for this because the import statement is more than 160 lines and Checkstyle won't let you put a line break in an import statement. Sigh.

YetAnotherMatt: Maybe `IdAndEligibilityResponseTestDataFactory`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementDecisionFactoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+        //When
+        EligibilityAndEntitlementDecision eligibilityAndEntitlementDecision = factory.buildDecision(eligibilityResponse, entitlement, Optional.empty(), true);
+
+        //Then
+        EligibilityAndEntitlementDecision expectedDecision = aValidDecisionBuilder()
+                .identityAndEligibilityResponse(eligibilityResponse)
+                .hmrcHouseholdIdentifier(null)
+                .voucherEntitlement(null)
+                .eligibilityStatus(EligibilityStatus.DUPLICATE)
```
YetAnotherMatt: I think we should extract the `isDuplicate` parameter out to a variable to make it clearer. Took me a while to understand why we expected a duplicate status.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementDecisionFactoryTest.java
mikeSmith08: Yes, that works, the static import line is now 159 chars, so perfect! :-)

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementDecisionFactoryTest.java
mikeSmith08: Yes, good idea, will do.

PR #430, opened 2019-11-21
HTBHF-2641 Added Processor for SaveNewCard message type.

dwybourn:
Added a new processor to handle messages of type SAVE_NEW_CARD. The logic of creating a payment cycle and sending a MAKE_FIRST_PAYMENT message is currently duplicated in RequestNewCardProcesser. This is ok as currently nothing is creating a SAVE_NEW_CARD message. In the next PR the RequestNewCardProcesser will be stripped down to just request the new card then create a SAVE_NEW_CARD message.

The audit new card event was changed to take a cardAccountId rather than a full card response as only the card account id exists on the SAVE_NEW_CARD message.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
@@ -2,6 +2,7 @@ package uk.gov.dhsc.htbhf.claimant.message;
 
 public enum MessageType {
     REQUEST_NEW_CARD,
+    SAVE_NEW_CARD,
     MAKE_FIRST_PAYMENT,
     MAKE_PAYMENT,
     SEND_EMAIL,
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
@@ -2,7 +2,7 @@ package uk.gov.dhsc.htbhf.claimant.message;
 
 public enum MessageType {
     REQUEST_NEW_CARD,
-    SAVE_NEW_CARD,
+    COMPLETE_NEW_CARD_PROCESS,
     MAKE_FIRST_PAYMENT,
     MAKE_PAYMENT,
     SEND_EMAIL,
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
@@ -1,0 +1,85 @@
+package uk.gov.dhsc.htbhf.claimant.message.processor;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import uk.gov.dhsc.htbhf.claimant.entity.Claim;
+import uk.gov.dhsc.htbhf.claimant.entity.Message;
+import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
+import uk.gov.dhsc.htbhf.claimant.message.MessageQueueClient;
+import uk.gov.dhsc.htbhf.claimant.message.MessageStatus;
+import uk.gov.dhsc.htbhf.claimant.message.MessageType;
+import uk.gov.dhsc.htbhf.claimant.message.MessageTypeProcessor;
+import uk.gov.dhsc.htbhf.claimant.message.context.MessageContextLoader;
+import uk.gov.dhsc.htbhf.claimant.message.context.SaveNewCardMessageContext;
+import uk.gov.dhsc.htbhf.claimant.message.payload.MakePaymentMessagePayload;
+import uk.gov.dhsc.htbhf.claimant.repository.ClaimRepository;
+import uk.gov.dhsc.htbhf.claimant.service.ClaimMessageSender;
+import uk.gov.dhsc.htbhf.claimant.service.audit.EventAuditor;
+import uk.gov.dhsc.htbhf.claimant.service.payments.PaymentCycleService;
+
+import javax.transaction.Transactional;
+
+import static uk.gov.dhsc.htbhf.claimant.message.MessagePayloadFactory.buildMakePaymentMessagePayload;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageStatus.COMPLETED;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageType.MAKE_FIRST_PAYMENT;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageType.SAVE_NEW_CARD;
+import static uk.gov.dhsc.htbhf.claimant.model.ClaimStatus.ACTIVE;
+import static uk.gov.dhsc.htbhf.claimant.reporting.ClaimAction.UPDATED_FROM_NEW_TO_ACTIVE;
+
+/**
+ * Responsible for processing {@link MessageType#SAVE_NEW_CARD} messages by:
+ * Saving the card account id to the claim,
+ * Creating a PaymentCycle for the claim,
+ * Sending a {@link MessageType#MAKE_FIRST_PAYMENT} message.
+ */
+@Component
+@AllArgsConstructor
+@Slf4j
+public class SaveNewCardMessageProcessor implements MessageTypeProcessor {
+
+    private MessageContextLoader messageContextLoader;
+    private ClaimRepository claimRepository;
+    private PaymentCycleService paymentCycleService;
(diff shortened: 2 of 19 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+import static uk.gov.dhsc.htbhf.claimant.message.MessagePayloadFactory.buildMakePaymentMessagePayload;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageStatus.COMPLETED;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageType.MAKE_FIRST_PAYMENT;
+import static uk.gov.dhsc.htbhf.claimant.message.MessageType.SAVE_NEW_CARD;
+import static uk.gov.dhsc.htbhf.claimant.model.ClaimStatus.ACTIVE;
+import static uk.gov.dhsc.htbhf.claimant.reporting.ClaimAction.UPDATED_FROM_NEW_TO_ACTIVE;
+
+/**
+ * Responsible for processing {@link MessageType#SAVE_NEW_CARD} messages by:
+ * Saving the card account id to the claim,,
```
YetAnotherMatt: Trivial, but there's an extra comma here

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/EventAuditor.java
```diff
@@ -58,17 +57,17 @@ public void auditUpdatedClaim(Claim claim, List<UpdatableClaimantField> updatedF
...
-     * @param cardResponse The card response which must not be null
+     * @param claimId      The claim id which must not be null
+     * @param cardAccountId The card account id which must not be null
      */
-    public void auditNewCard(UUID claimId, CardResponse cardResponse) {
-        if (claimId == null || cardResponse == null) {
-            log.warn("Unable to audit new card event with claimId: {} and cardResponse: {}. Both fields must not be null", claimId, cardResponse);
+    public void auditNewCard(UUID claimId, String cardAccountId) {
+        if (claimId == null || cardAccountId == null) {
+            log.warn("Unable to audit new card event with claimId: {} and cardAccountId: {}. Both fields must not be null", claimId, cardAccountId);
```
YetAnotherMatt: I know this isn't wording changed in this PR, but if you happen to be making changes anyway: can we replace "Both fields must not be null" with "Both fields must be provided"

dwybourn: Updated.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
dwybourn: Removed extra comma.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+    public MessageType supportsMessageType() {
+        return SAVE_NEW_CARD;
+    }
+
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public MessageStatus processMessage(Message message) {
+        SaveNewCardMessageContext context = messageContextLoader.loadSaveNewCardContext(message);
+        saveCardAccountIdToClaim(context);
+        eventAuditor.auditNewCard(context.getClaim().getId(), context.getCardAccountId());
```
mikeSmith08: Shouldn't this call now be `eventAuditor.auditSaveCard`?

mikeSmith08:
I'm not sure I'm too happy with the name of this processor, firstlt its not saving the card to the database, its saving the card id, and secondly its doing a number of things like creating the new PaymentCycle, so its doing many things. Not sure I have a good alternative, but I still keep looking at it and have to concentrate on the difference between the NewCardMessageProcessor and the SaveCardMessageProcessor.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
```diff
@@ -2,6 +2,7 @@
 
 public enum MessageType {
     REQUEST_NEW_CARD,
+    SAVE_NEW_CARD,
```
mikeSmith08: Should this be SAVE_NEW_CARD_ID?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/SaveNewCardMessageProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private PaymentCycle createAndSavePaymentCycle(SaveNewCardMessageContext context) {
+        Claim claim = context.getClaim();
+        return paymentCycleService.createAndSavePaymentCycleForEligibleClaim(
+                claim,
+                claim.getClaimStatusTimestamp().toLocalDate(),
+                context.getEligibilityAndEntitlementDecision());
+    }
+
+    private void saveCardAccountIdToClaim(SaveNewCardMessageContext context) {
```
mikeSmith08: It would be better to have the Claim being passed as parameter to this method.

mikeSmith08: Actually, should this method be in another class, maybe ClaimService?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/audit/EventAuditor.java
```diff
@@ -58,17 +57,17 @@ public void auditUpdatedClaim(Claim claim, List<UpdatableClaimantField> updatedF
...
-     * @param claimId      The claim id
-     * @param cardResponse The card response which must not be null
+     * @param claimId      The claim id which must not be null
+     * @param cardAccountId The card account id which must not be null
      */
-    public void auditNewCard(UUID claimId, CardResponse cardResponse) {
-        if (claimId == null || cardResponse == null) {
-            log.warn("Unable to audit new card event with claimId: {} and cardResponse: {}. Both fields must not be null", claimId, cardResponse);
+    public void auditNewCard(UUID claimId, String cardAccountId) {
+        if (claimId == null || cardAccountId == null) {
```
mikeSmith08: Is it worth being a bit more strict on the cardAccountId and using `StringUtils.isBlank()`. I guess the question is whether or not a cardAccountId which is a blank string is acceptable or not.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/RequestNewCardServiceTest.java
```diff
@@ -69,7 +69,7 @@ void shouldCallCardClientAndUpdateClaim() {
 
         verify(cardRequestFactory).createCardRequest(claim);
         verify(cardClient).requestNewCard(cardRequest);
-        verify(eventAuditor).auditNewCard(claim.getId(), cardResponse);
+        verify(eventAuditor).auditNewCard(claim.getId(), cardResponse.getCardAccountId());
```
mikeSmith08: Can we use the actual card account id here rather than the method please?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/audit/EventAuditorTest.java
```diff
@@ -105,10 +104,10 @@ void shouldLogUpdatedClaimEvent() {
     void shouldLogEventForNewCard() {
         //Given
         UUID claimId = UUID.randomUUID();
-        CardResponse cardResponse = aCardResponse();
+        String cardAccountId = CARD_ACCOUNT_ID;
```
mikeSmith08: Why the need for the local variable, why not just use the constant each time?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/MessagePayloadTestDataFactory.java
```diff
@@ -45,8 +45,8 @@ public static RequestNewCardMessagePayload aNewCardRequestMessagePayload(UUID cl
                 .build();
     }
 
-    public static SaveNewCardMessagePayload aSaveCardMessagePayload(UUID claimId) {
-        return SaveNewCardMessagePayload.builder()
+    public static CompleteNewCardMessagePayload aSaveCardMessagePayload(UUID claimId) {
```
mikeSmith08: You need to change the method name here too, should be `aCompleteNewCardMessagePayload`

PR #435, opened 2019-11-25
HTBHF-2570 Create v2 of EligibilityAndEntitlementService and test.

mikeSmith08:
This currently isn't used anywhere, will be manually wired into the v2 controller once that controller has been created.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/v2/EligibilityAndEntitlementServiceV2.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/v2/EligibilityAndEntitlementServiceV2.java
@@ -1,0 +1,83 @@
+package uk.gov.dhsc.htbhf.claimant.service.v2;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleEntitlementCalculator;
+import uk.gov.dhsc.htbhf.claimant.entitlement.PaymentCycleVoucherEntitlement;
+import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
+import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
+import uk.gov.dhsc.htbhf.claimant.model.eligibility.EligibilityAndEntitlementDecision;
+import uk.gov.dhsc.htbhf.claimant.repository.ClaimRepository;
+import uk.gov.dhsc.htbhf.claimant.service.DuplicateClaimChecker;
+import uk.gov.dhsc.htbhf.claimant.service.EligibilityAndEntitlementDecisionFactory;
+import uk.gov.dhsc.htbhf.claimant.service.EligibilityAndEntitlementService;
+import uk.gov.dhsc.htbhf.dwp.model.v2.IdentityAndEligibilityResponse;
+
+import java.time.LocalDate;
+import java.util.Optional;
+import java.util.UUID;
+
+@AllArgsConstructor
+@Slf4j
+public class EligibilityAndEntitlementServiceV2 implements EligibilityAndEntitlementService {
+
+    private final EligibilityClientV2 client;
+    private final DuplicateClaimChecker duplicateClaimChecker;
+    private final ClaimRepository claimRepository;
+    private final PaymentCycleEntitlementCalculator paymentCycleEntitlementCalculator;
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/v2/EligibilityAndEntitlementServiceV2.java
```diff
@@ -0,0 +1,85 @@
...
+import uk.gov.dhsc.htbhf.claimant.service.EligibilityAndEntitlementDecisionFactory;
+import uk.gov.dhsc.htbhf.claimant.service.EligibilityAndEntitlementService;
+import uk.gov.dhsc.htbhf.dwp.model.v2.IdentityAndEligibilityResponse;
+
+import java.time.LocalDate;
+import java.util.Optional;
+import java.util.UUID;
+
+@AllArgsConstructor
+@Slf4j
```
dwybourn: Needs to be a component. Can be done in this or next PR.

mikeSmith08: No, it doesn't, there are now two implementations of the interface which means that we need to control which instance goes where. In the final PR for this story I will be using @Configuration beans to setup the v1 and v2 controllers and required services. Either way, not for this PR.

YetAnotherMatt: I think we're going to have to wire these in as `@Bean`s and make V1 the primary.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/v2/EligibilityAndEntitlementServiceV2.java
```diff
@@ -0,0 +1,85 @@
...
+     * the external eligibility service is called.
+     * Claimants determined to be eligible by the external eligibility service must still either be pregnant or have children under 4,
+     * otherwise they will be ineligible.
+     *
+     * @param claimant the claimant to check the eligibility for
+     * @return the eligibility and entitlement for the claimant
+     */
+    @Override
+    public EligibilityAndEntitlementDecision evaluateNewClaimant(Claimant claimant) {
+        log.debug("Looking for live claims for the given NINO");
```
dwybourn: I was wondering if we could move some of these log statements into the called methods so that this method is smaller? Not saying we should, something to consider.

mikeSmith08: I can't on the first one a its a repository interface, but I can for the other two.

mikeSmith08: Done of the two that I can move.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/v2/EligibilityAndEntitlementServiceV2.java
```diff
@@ -0,0 +1,83 @@
...
+        Optional<UUID> liveClaimsWithNino = claimRepository.findLiveClaimWithNino(claimant.getNino());
+        IdentityAndEligibilityResponse identityAndEligibilityResponse = client.checkIdentityAndEligibility(claimant);
+        PaymentCycleVoucherEntitlement entitlement = paymentCycleEntitlementCalculator.calculateEntitlement(
+                Optional.ofNullable(claimant.getExpectedDeliveryDate()),
+                identityAndEligibilityResponse.getDobOfChildrenUnder4(),
+                LocalDate.now());
+        if (liveClaimsWithNino.isPresent()) {
+            return eligibilityAndEntitlementDecisionFactory.buildDecision(identityAndEligibilityResponse,
```
dwybourn: We should probably create a new method in the factory. The last parameters are null and false which you can't tell the relevance of unless you look at the buidDecision signature.

mikeSmith08: These values are only temporary until we do the follow-up story, then they will have values, so don't really want to add a method for it to be subsequently removed. If you remember, we did pair on this together last week and the extra methods in the factory didn't work well and we decided against them anyway.

PR #450, opened 2019-11-28
HTBHF-2672 Create v3 claim, claimant and address dto objects and test.

mikeSmith08:
Not currently used but will replace the existing objects in the next PR for the v3 controller.

The key changes in the new versions are that the regex for NINO and postcode have changed to match the DWP spec, apart from that the DTOs, test data factories and unit tests are the same as the existing ones.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/v3/AddressDTOV3.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.validation.constraints.Size;
+
+import static uk.gov.dhsc.htbhf.claimant.regex.PostcodeRegex.CHANNEL_ISLAND_POST_CODE_REGEX;
+import static uk.gov.dhsc.htbhf.claimant.regex.PostcodeRegex.UK_POST_CODE_REGEX_V3;
+
+@Data
+@Builder
+@AllArgsConstructor(onConstructor_ = {@JsonCreator})
+@ApiModel(description = "Multi purpose address object")
+public class AddressDTOV3 {
```
YetAnotherMatt: Does this differ from the existing addressDTO in the model package?

mikeSmith08: Yes, the postcode regex has changed.

PR #453, opened 2019-11-28
HTBHF-2673 Store identity and eligibility response from initial claim …

mikeSmith08:
…in the database.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claim.java
```diff
@@ -74,6 +75,10 @@
     @Setter(AccessLevel.PRIVATE)
     private LocalDateTime cardStatusTimestamp;
 
+    @Column(name = "initial_identity_and_eligibility_response")
+    @Type(type = "json")
+    private IdentityAndEligibilityResponse initialIdentityAndEligibilityResponse;
```
YetAnotherMatt: I just remembered that this will need to change to be the new identityAndEligibilityResponse that includes both household identifiers when we do the work on HTBHF-2654. Maybe we should shelve this until then?

mikeSmith08: Yes, a very fair point, I'll park this then until we do the HMRC story which is up next.

PR #456, opened 2019-11-29
Feature/reactive pending expiry claim

dwybourn:
When an eligible response comes back for a claim in PENDING_EXPIRY, that claim and card status are moved back to ACTIVE.
A payment restarted email is then sent, this replaces the regular payment email.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/eligibility/EligibilityDecisionHandlerTest.java
```diff
@@ -318,9 +334,9 @@ void shouldExpirePendingExpiryClaim() {
 
     private void verifyClaimSavedWithStatus(ClaimStatus claimStatus, CardStatus cardStatus) {
         ArgumentCaptor<Claim> argumentCaptor = ArgumentCaptor.forClass(Claim.class);
-        verify(claimRepository, times(2)).save(argumentCaptor.capture());
```
dwybourn: claimRepository.save() is only called once now with an updated claim and card status.

PR #461, opened 2019-12-02
HTBHF-2677 Update to latest java-common and use latest TestConstants …

mikeSmith08:
…and test data factories.

This touched quite a lot of files so putting this up as its own PR first before we update to use the CombinedIdentityAndEligibilityResponse.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/creator/ClaimantLoader.java
```diff
@@ -154,7 +153,7 @@ private PaymentCycle createPaymentCycleEndingYesterday(Claim claim, List<ChildIn
                 .voucherEntitlement(voucherEntitlement)
                 .eligibilityStatus(EligibilityStatus.ELIGIBLE)
                 .identityAndEligibilityResponse(
-                        anIdentityMatchedEligibilityConfirmedUCResponseWithAllMatches(VerificationOutcome.NOT_SUPPLIED, childrenDatesOfBirth))
+                        anIdentityMatchedEligibilityConfirmedUCResponseWithAllMatches(childrenDatesOfBirth))
```
YetAnotherMatt: what happened to the VerificationOutcome - is NOT_SUPPLIED the default?

mikeSmith08: Yes, that was done in the previous java-common PR, almost all usages were changing it to NOT_SUPPLIED so it seemed sensible to make it the default in the factory.

PR #464, opened 2019-12-03
HTBHF-2677 Change to use CombinedIdentityAndEligibilityResponse from …

mikeSmith08:
…eligibility-service rather than the DWP IdentityAndEligibilityResponse

Sorry for this size of this PR (although net number of lines changed is negative :-) ) but this change had to ripple through the whole of the claimant service.

Some tests needed to be removed because the scenarios they were testing (with/without HMRC household identifier) are no longer applicable because this gets taken directly from the response from the eligibility-service response and there is no difference between v2 and v3 versions in that respect.

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleServiceTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleServiceTest.java
@@ -14,7 +14,8 @@ import uk.gov.dhsc.htbhf.claimant.model.ClaimStatus;
 import uk.gov.dhsc.htbhf.claimant.model.eligibility.EligibilityAndEntitlementDecision;
 import uk.gov.dhsc.htbhf.claimant.repository.PaymentCycleRepository;
 import uk.gov.dhsc.htbhf.claimant.testsupport.EligibilityAndEntitlementTestDataFactory;
-import uk.gov.dhsc.htbhf.dwp.model.v2.IdentityAndEligibilityResponse;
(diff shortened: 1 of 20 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentCycleServiceTest.java
```diff
@@ -115,11 +116,14 @@ void shouldCreateNewPaymentCycleWithoutExpectedDueDateForEligibleClaimWithoutPre
...
         PaymentCycleVoucherEntitlement entitlement = aPaymentCycleVoucherEntitlementWithoutPregnancyVouchers();
+        CombinedIdentityAndEligibilityResponse identityAndEligibilityResponse = CombinedIdentityAndEligibilityResponseTestDataFactory
+                .anIdentityMatchedEligibilityConfirmedUCResponseWithAllMatches(datesOfBirth);
         EligibilityAndEntitlementDecision decision = EligibilityAndEntitlementTestDataFactory.aValidDecisionBuilder()
                 .existingClaimId(claim.getId())
                 .voucherEntitlement(entitlement)
                 .dateOfBirthOfChildren(datesOfBirth)
-                .identityAndEligibilityResponse(anIdentityMatchedEligibilityConfirmedUCResponseWithAllMatches(datesOfBirth))
+                .identityAndEligibilityResponse(
+                        identityAndEligibilityResponse)
```
YetAnotherMatt: this could be on one line, but it's not worth amending this PR for

mikeSmith08: Yep, I'll do it in the next PR.

PR #470, opened 2019-12-04
NO-JIRA Add missing unit test for ClaimControllerV3.

mikeSmith08:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/v3/ClaimControllerV3.java
```diff
@@ -48,7 +48,6 @@ public ClaimControllerV3(@Qualifier("v3ClaimService") ClaimService v3ClaimServic
         return claimController.createOrUpdateClaim(buildClaimDTO(claimDTOV3));
     }
 
-    //TODO MRS 28/11/2019: BeanUtils copyProperties possibly.
```
mikeSmith08: I initially wrote the unit test below because I wanted to replace the code in this controller with BeanUtils to copy the fields, but as BeanUtils only does a shallow copy, I had to manually copy each sub-object and reconstruct the whole object, and that code was just as long and pretty horrible as I had to create a no-args constructor and use setters for the sub-objects. Hence decided to ditch this in the interim and revisit this when we decommission the v2 controller.

PR #473, opened 2019-12-06
HTBHF-2525 Additional reporting tests.

dwybourn:
- Updated reporting tests to assert on the number of children. 
- Added new lifecycle integration test for loss of child. This test asserts that we report the most recent children received from the eligibility service, and not the children from the initial claim request.
- Updated existing lifecycle integration test to assert on report claim and payment events.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/ReportClaimIntegrationTest.java
```diff
@@ -55,25 +56,25 @@ void shouldReportNewClaimToGoogleAnalyticsWithPostcodeData() throws JsonProcessi
...
         claimRepository.save(claim);
         String postcode = claim.getClaimant().getAddress().getPostcode();
         stubPostcodesIoAndGoogleAnalytics(postcode);
 
-        claimMessageSender.sendReportClaimMessage(claim, ONE_CHILD_UNDER_ONE_AND_ONE_CHILD_BETWEEN_ONE_AND_FOUR, ClaimAction.REJECTED);
+        claimMessageSender.sendReportClaimMessage(claim, ONE_CHILD_UNDER_ONE_AND_ONE_CHILD_BETWEEN_ONE_AND_FOUR, REJECTED);
         messageProcessorScheduler.processReportClaimMessages();
 
         wiremockManager.verifyPostcodesIoCalled(postcode);
-        wiremockManager.verifyGoogleAnalyticsCalledForClaimEvent(claim, ClaimAction.REJECTED, trackingId);
```
dwybourn: Switched around the ClaimStatus and ClaimAction static import so that this line would be under 160 characters.

PR #476, opened 2019-12-06
HTBHF-2702 Store current identity & eligibility response in claim

adeel718:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/service/claim/ClaimService.java
```diff
@@ -145,6 +145,7 @@ private Claim buildClaim(ClaimRequest claimRequest, EligibilityAndEntitlementDec
                 .deviceFingerprintHash(fingerprintHash)
                 .webUIVersion(claimRequest.getWebUIVersion())
                 .initialIdentityAndEligibilityResponse(decision.getIdentityAndEligibilityResponse())
+                .currentIdentityAndEligibilityResponse(decision.getIdentityAndEligibilityResponse())
```
mikeSmith08: This only deals with the current identity and eligibility decision being set when the claim is first made, are you going to do the change of identity and eligibility response for a new Payment Cycle in another PR?

adeel718: I'm going to raise another PR for this. Am I correct that this will be in method 'createAndSavePaymentCycleForEligibleClaim' within the 'PaymentCycleService'?

mikeSmith08: No, its done by the `DetermineEntitlementMessageProcessor` as a part of the payment cycle processing, so when it stores the `IdentityAndEligibilityResponse` on the payment cycle, it will also need to store it on the Claim.

YetAnotherMatt: I think we'll actually put the update in the `DetermineEntitlementMessageProcessor.processMessage` method - just after updating the paymentCycle

PR #478, opened 2019-12-09
HTBHF-2702 Update current identityand eligibility on payment cycle

adeel718:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
```diff
@@ -75,6 +79,7 @@ public MessageStatus processMessage(Message message) {
                 previousPaymentCycle);
 
         paymentCycleService.updatePaymentCycle(currentPaymentCycle, decision);
+        claimService.updateCurrentIdentityAndEligibilityResponse(claim, decision.getIdentityAndEligibilityResponse());
```
YetAnotherMatt: I'm thinking that we should maybe refactor these lines into another component, as this message processor has 6 constructor args again
```
        paymentCycleService.updatePaymentCycle(currentPaymentCycle, decision);
        claimService.updateCurrentIdentityAndEligibilityResponse(claim, decision.getIdentityAndEligibilityResponse());
        handleDecision(claim, previousPaymentCycle, currentPaymentCycle, decision, message.getCreatedTimestamp());
        if (claim.getClaimStatus() == PENDING_EXPIRY) {
            paymentCycleService.updateEndDateForClaimBecomingPendingExpiry(currentPaymentCycle);
        }
```
Only problem is, I can't think what to call it. Thoughts?
Either way, it wouldn't be part of this PR

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -67,6 +67,8 @@
     private static final int CARD_BALANCE_IN_PENCE_BEFORE_DEPOSIT = 88;
     private static final LocalDate NOT_PREGNANT = null;
     private static final String CARD_ACCOUNT_ID = UUID.randomUUID().toString();
+    
+    // TODO HTBHF-2475 - update integration tests to confirm currentIdentityAndEligibilityResponse is updated, once we are calling v3 eligibility service
```
mikeSmith08: Shouldn't this be a part of this PR seeing as the functionality is being added, otherwise it isn't integration tested?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/DetermineEntitlementMessageProcessor.java
mikeSmith08: I do agree, but bear in mind that one of the dependencies is a config value, so technically its 5 dependencies, so maybe not so desperate. Might be worth a look anyway though.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/claim/ClaimServiceTest.java
```diff
@@ -520,6 +522,23 @@ void shouldSaveClaimantWhenEligibilityThrowsException() {
         assertThat(eventCaptor.getValue().getEventMetadata().get(FailureEvent.FAILED_EVENT_KEY)).isEqualTo(ClaimEventType.NEW_CLAIM);
         verifyNoInteractions(claimMessageSender);
     }
+    
+    @Test
+    void shouldUpdateClaimWithCurrentIdentityAndEligibilityResponse() {
+        // given
+        Claim claim = aValidClaim();
+        CombinedIdentityAndEligibilityResponse response =
```
mikeSmith08: Any particular reason you didn't use a test data factory method here? It can be any `CombinedIdentityAndEligibilityResponse` so seems like using one of the methods from there would be fine.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
YetAnotherMatt: We can't add that check until we've updated the scheduled services to use v3, at least not easily

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/claim/ClaimServiceTest.java
YetAnotherMatt: only because we couldn't find it in eclipse ('Identify' was shortened to 'Id' in the factory name). We've used the factory now

PR #479, opened 2019-12-09
HTBHF-2699 send qualifying benefit to ga

adeel718:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
```diff
@@ -79,17 +80,21 @@ protected Long calculateQueueTime(LocalDateTime timestamp) {
     protected Map<String, Object> createCommonCustomDimensions(ReportEventMessageContext context) {
         Map<String, Object> customDimensions = new TreeMap<>();
         customDimensions.put(USER_TYPE.getFieldName(), ONLINE.name());
+        Claim claim = context.getClaim();
         ClaimantCategory claimantCategory = claimantCategoryCalculator
-                .determineClaimantCategory(context.getClaim().getClaimant(), context.getDatesOfBirthOfChildren(), context.getTimestamp().toLocalDate());
+                .determineClaimantCategory(claim.getClaimant(), context.getDatesOfBirthOfChildren(), context.getTimestamp().toLocalDate());
```
mikeSmith08: We can now get the dates of birth of children from the current IdentityAndEligibilityResponse on the claim so do we need these dates of birth on the context anymore? If we are going to do that then not using them here would be the first easy step in that process.

YetAnotherMatt: There will be a separate PR to remove the dates of birth

YetAnotherMatt: After a discussion with David, who pointed out that as our messages are asynchronous it's possible that the claim could have been updated since the message was created. For that reason we should keep the children's dob in the context.

YetAnotherMatt: What we'll do instead is replace the list of children's dob with the complete CombinedIdentityAndEligibilityResponse - in a separate PR

PR #484, opened 2019-12-10
HTBHF-2705 Updated determine entitlement processor to use v2 eligibil…

dwybourn:
Updated determine entitlement processor to use v2 eligibility endpoint and claim lifecycle tests to use v3 claim endpoint.
The plan was to just update the entitlement processor, but the changes to common test classes meant it easier (and less code) to switch over the claim lifecycle tests to v3 of claim at the same time.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/PaymentCycleIntegrationTests.java
```diff
@@ -365,12 +366,12 @@ void shouldTestClaimantBecomingIneligible(List<LocalDate> previousCycleChildrenD
     @DisplayName("Integration test for HTBHF-2182 status set to Expired and no email sent to Claimant who has no children in current cycle, "
             + "not pregnant, children over 4 present in previous cycle, response from DWP has no children")
     @ParameterizedTest(name = "DWP eligibility status={0}")
-    @ValueSource(strings = {"ELIGIBLE", "INELIGIBLE"})
-    void shouldTestClaimBecomingExpiredWhenRollingOffTheScheme(EligibilityStatus eligibilityStatus) throws JsonProcessingException {
+    @EnumSource(EligibilityOutcome.class)
+    void shouldTestClaimBecomingExpiredWhenRollingOffTheScheme(EligibilityOutcome eligibilityOutcome) throws JsonProcessingException {
```
mikeSmith08: I think parameterising these tests just by `EligibilityOutcome` isn't correct as it only tells part of the story, it should be based on the whole or the `IdentityAndEligibilityResponse` as the name suggests, its not just the eligibility that is being returned and relevant here. Let's run through these tests because I don't think this is correct like this.

PR #494, opened 2019-12-12
HTBHF-2705 Renamed buildClaimRequestEntityForUri to buildClaimRequest…

dwybourn:
…Entity and changed method to use v3 claim endpoint instead of provided parameter.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceAssertionUtils.java
```diff
@@ -36,10 +39,10 @@ public static void assertClaimantMatchesClaimantDTO(ClaimantDTO claimant, Claima
     }
 
     // TODO HTBHF-2705 Change this to hardcoded v3 claim endpoint instead of a URI parameter.
```
YetAnotherMatt: probably want to remove this comment too

dwybourn: Removed.

PR #496, opened 2019-12-12
NOJIRA Refactored buildDecision method to not take in an existingClai…

dwybourn:
…mId as this was always being set to null in the production code.

### DIFF
--- a/api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
+++ b/api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
@@ -89,7 +89,7 @@ class EligibilityAndEntitlementServiceTest {
 
         //Then
         assertThat(result).isEqualTo(decisionResponse);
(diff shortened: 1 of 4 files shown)


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -121,19 +121,18 @@ private void setupCommonMocks(UUID existingClaimId) {
         given(claimRepository.findLiveClaimWithNino(any())).willReturn(Optional.ofNullable(existingClaimId));
     }
 
-    private void verifyDecisionFactoryCalledCorrectly(UUID existingClaimId, boolean duplicate) {
-        verify(eligibilityAndEntitlementDecisionFactory).buildDecision(IDENTITY_AND_ELIGIBILITY_RESPONSE,
-                VOUCHER_ENTITLEMENT, existingClaimId, duplicate);
+    private void verifyDecisionFactoryCalledCorrectly(boolean duplicate) {
```
mikeSmith08: Does this method really need to exist now, or could/should it be inlined?

PR #498, opened 2019-12-12
HTBHF-2727 Added tests for matching on children's date of birth.

dwybourn:
Also added cleanup step for existing integration tests that were causing the build  to fail as new claims were being rejected as a duplicate.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/ChildrenDateOfBirthMatchIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        ResponseEntity<ClaimResultDTO> response = restTemplate.exchange(buildClaimRequestEntity(claim), ClaimResultDTO.class);
+
+        assertThat(response.getStatusCode()).isEqualTo(CREATED);
+        assertThat(response.getBody().getClaimStatus()).isEqualTo(NEW);
+    }
+
+    // Provides arguments for successfully creating a new claim. Arguments are: expected delivery date, declared children dates of birth
+    // and children dates of birth returned from benefit agency.
+    private static Stream<Arguments> providedArgumentsForCreateNewClaim() {
```
YetAnotherMatt: Can we rename this to something like `datesOfBirthForValidClaims`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ChildrenDateOfBirthMatchIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @AfterEach
+    void cleanup() {
+        wiremockManager.stopWireMock();
+        repositoryMediator.deleteAllEntities();
+    }
+
+    @ParameterizedTest
+    @MethodSource("providedArgumentsForCreateNewClaim")
+    void shouldCreateNewClaim(LocalDate expectedDeliveryDate,
```
YetAnotherMatt: `shouldCreateNewClaimWhenPregnantOrChildDatesOfBirthMatch`

mikeSmith08: I would argue whether we even need this as an extra integration test as its just the happy, surely we just need to add `assertThat(response.getBody().getVerificationResult().getIsPregnantOrAtLeast1ChildMatched()).isTrue();` to the existing happy path integration test we just have the one negative test.

dwybourn: I guess it depends if we want an integration test for each decision path. Personally I'm in favour of more integration tests.

YetAnotherMatt: One thing to bear in mind - the tests take 8 minutes or more on Adeel's machine.

mikeSmith08: There are certainly pros and cons for each side of this argument, I don't feel too strongly either way so happy if it stays.

PR #499, opened 2019-12-13
  HTBHF-2727 New claim requests are now rejected if the claimant is n…

dwybourn:
…ot pregnant and none of their declared children dates of birth match that from the benefit agency.


api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/VerificationResultFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class VerificationResultFactory {
+
+    /**
+     * Creates a {@link VerificationResult} from a given {@link Claimant} and {@link CombinedIdentityAndEligibilityResponse}.
+     * @param claimant claimant details
+     * @param identityAndEligibilityResponse response from eligibility serivce
+     * @return a {@link VerificationResult}
+     */
+    public static VerificationResult buildVerificationResult(Claimant claimant, CombinedIdentityAndEligibilityResponse identityAndEligibilityResponse) {
```
dwybourn: All of this code is lifted and shifted from ClaimResultDTO, same goes for the unit test.

PR #506, opened 2019-12-17
HTBHF-2470 Send update your address letter when eligible claimants address mismatches

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/VerificationResult.java
```diff
@@ -36,7 +36,8 @@
 
     @JsonIgnore
     public boolean isAddressMismatch() {
-        return VerificationOutcome.NOT_MATCHED == addressLine1Match || VerificationOutcome.NOT_MATCHED == postcodeMatch;
+        return EligibilityOutcome.CONFIRMED == eligibilityOutcome
+                && (VerificationOutcome.NOT_MATCHED == addressLine1Match || VerificationOutcome.NOT_MATCHED == postcodeMatch);
```
YetAnotherMatt: I checked with Karen, and she wanted to include a check that we would only send an email & letter for an address mismatch if the claimant was otherwise eligible.

mikeSmith08: I think this is a redundant check though because if the EligibilityOutcome is not CONFIRMED then these values will always be NOT_SET, I think this should be removed as it confuses the basis of how the VerificationResult is built.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/VerificationResultTestDataFactory.java
```diff
@@ -71,6 +71,14 @@ public static VerificationResult aFullAddressMismatchVerificationResult() {
                 .build();
     }
 
+    public static VerificationResult anIneligibleFullAddressMismatchVerificationResult() {
```
mikeSmith08: I disagree with this VerificationResult. As I said above, this is an invalid state to be in, if the EligibilityOutcome is NOT_CONFIRMED then all VerificationOutcomes will be NOT_SET.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/VerificationResultTest.java
```diff
@@ -37,6 +37,12 @@ void shouldHaveAddressMismatchForFullAddressMismatch() {
         assertThat(result.isAddressMismatch()).isTrue();
     }
 
+    @Test
+    void shouldNotHaveAddressMismatchWhenNotEligible() {
```
mikeSmith08: This test should be removed too as the VerificationResult should not be able to get in this state.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/ClaimMessageSenderTest.java
```diff
@@ -159,4 +157,13 @@ void shouldSendDecisionPendingMessage() {
         verify(messageQueueClient).sendMessage(emailMessagePayload, SEND_EMAIL);
     }
 
+    @Test
+    void shouldSendUpdateYourAddressLetterMessage() {
+        Claim claim  = aValidClaim();
+
+        claimMessageSender.sendUpdateYourAddressLetterMessage(claim);
+
+        LetterMessagePayload payload = LetterMessagePayloadFactory.buildLetterPayloadWithAddressOnly(claim, LetterType.UPDATE_YOUR_ADDRESS);
```
mikeSmith08: Can we call this `expectedPayload`?

PR #516, opened 2020-01-07
AFHS-542 Added a childrenUnderFourPresentAtEndOfCycle flag to NextPaymentCycleSummary

YetAnotherMatt:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/ChildDateOfBirthCalculator.java
```diff
@@ -65,9 +65,11 @@ public NextPaymentCycleSummary getNextPaymentCycleSummary(PaymentCycle paymentCy
         LocalDate lastEntitlementDateInNextCycle = getLatestEntitlementDateFromCycleStartDate(nextCycleStartDate);
         int childrenAgedOneAffectingNextPayment = countChildrenOfAge(paymentCycle, lastEntitlementDateInCurrentCycle, lastEntitlementDateInNextCycle, 1);
         int childrenAgedFourAffectingNextPayment = countChildrenOfAge(paymentCycle, lastEntitlementDateInCurrentCycle, lastEntitlementDateInNextCycle, 4);
+        boolean childrenUnderFourAtEndOfCycle = hadChildrenUnderFourAtGivenDate(paymentCycle.getChildrenDob(), lastEntitlementDateInNextCycle);
```
mikeSmith08: Shouldn't this be `hasChildrenUnderFourAtGivenDate` because the date is in the future?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/NextPaymentCycleSummary.java
```diff
@@ -14,11 +14,16 @@
 
     private int numberOfChildrenTurningOne;
     private int numberOfChildrenTurningFour;
+    private boolean childrenUnderFourPresentAtEndOfCycle;
```
mikeSmith08: It feels to me like we'll also need to store whether they are considered pregnant for the next cycle too otherwise the calculations will need to be made elsewhere when they could nicely be encapsulated here.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/ChildDateOfBirthCalculator.java
YetAnotherMatt: I'm re-using an existing method - where originally used the date is in the past. However it might be that 'has' is more generic and a better fit now.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/NextPaymentCycleSummary.java
YetAnotherMatt: I hadn't wanted to include pregnancy here as the summary object has only referenced children until now.

YetAnotherMatt: Adeel and I will look at it in the next PR and see where pregnancy fits best

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/ChildDateOfBirthCalculator.java
YetAnotherMatt: updated

PR #520, opened 2020-01-08
Send payment stopping email when youngest child is turning four

adeel718:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/PaymentCycleNotificationHandler.java
```diff
@@ -54,11 +58,22 @@ private boolean voucherEntitlementIndicatesNewChildFromPregnancy(PaymentCycle pa
 
     private void handleUpcomingBirthdayEmails(PaymentCycle paymentCycle) {
         NextPaymentCycleSummary nextPaymentCycleSummary = childDateOfBirthCalculator.getNextPaymentCycleSummary(paymentCycle);
-        if (nextPaymentCycleSummary.hasChildrenTurningFour()) {
-            upcomingBirthdayEmailHandler.sendChildTurnsFourEmail(paymentCycle, nextPaymentCycleSummary);
-        }
-        if (nextPaymentCycleSummary.hasChildrenTurningOne()) {
-            upcomingBirthdayEmailHandler.sendChildTurnsOneEmail(paymentCycle, nextPaymentCycleSummary);
+        boolean isNotPregnant = isNotPregnant(paymentCycle);
```
mikeSmith08: I don't think this adds anything making this a local variable as its not reused and looks nearly identical, we should inline this.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/PaymentCycleNotificationHandler.java
```diff
@@ -54,11 +58,22 @@ private boolean voucherEntitlementIndicatesNewChildFromPregnancy(PaymentCycle pa
...
+                upcomingBirthdayEmailHandler.sendChildTurnsFourEmail(paymentCycle, nextPaymentCycleSummary);
+            }
+            if (nextPaymentCycleSummary.hasChildrenTurningOne()) {
+                upcomingBirthdayEmailHandler.sendChildTurnsOneEmail(paymentCycle, nextPaymentCycleSummary);
+            }
         }
     }
-}
+
+    private boolean isNotPregnant(PaymentCycle paymentCycle) {
```
mikeSmith08: I think this should go in the `PregnancyEntitlementCalculator` and could have a better name than this, something like `isNotPregnantAfterCycle`.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/PaymentCycleNotificationHandler.java
YetAnotherMatt: true - it's a refactoring artefact I thought we'd removed again, oops.

YetAnotherMatt: Done

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/PaymentCycleNotificationHandler.java
YetAnotherMatt: Good point done - but the method is `claimantIsPregnantAfterCycle` to be in keeping with the existing `claimantIsPregnantInCycle`

PR #526, opened 2020-01-10
Added host:port, version and covert list LocalDates to List of String. 

AmarkumarJagtap:
It has added so swagger runs on local:8090 with default version 1.0 if not specified. Swagger was't working without any version, host:port and it was't able to translate list of DoBs to String list


api/src/main/java/uk/gov/dhsc/htbhf/claimant/ApiDocumentation.java
```diff
@@ -23,13 +29,19 @@
 @Slf4j
 public class ApiDocumentation {
 
-    @Value("${app.version:}") // use APP_VERSION env variable if available, otherwise give no version info
+    @Value("${app.version:1.0}") // use APP_VERSION env variable if available, otherwise give no version info
```
YetAnotherMatt: We need to update the comment: `// use APP_VERSION env variable if available, otherwise default to 1.0`

YetAnotherMatt: The reason for this change is that without a version number, the swagger isn't valid

mikeSmith08:
Would be good to have a description of why this change was needed, could you add a comment to the PR please as its not obvious from an initial look.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/ApiDocumentation.java
```diff
@@ -23,13 +29,19 @@
...
+    private TypeResolver typeResolver;
+
     @Bean
     public Docket apiDocket() {
         return new Docket(DocumentationType.SWAGGER_2)
-                .host("N/A")
+                .alternateTypeRules(
+                        newRule(typeResolver.resolve(List.class, LocalDate.class),
+                                typeResolver.resolve(List.class, String.class)))
+                .host("localhost:8090")
```
adeel718: This should probably be dynamic and get port number from config?

YetAnotherMatt: We thought about that and realised that whatever host/port number spring think's it's service from will almost certainly be hidden behinds a router somewhere - so this will only ever work on a local machine.

YetAnotherMatt:
> Would be good to have a description of why this change was needed, could you add a comment to the PR please as its not obvious from an initial look.

To quote from the new description: "Swagger was't working without any version, host:port and it was't able to translate list of DoBs to String list"

PR #528, opened 2020-01-13
added eligibilityOverrideOutcome field - AFHS-428

AmarkumarJagtap:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/NewClaimDTO.java
```diff
@@ -33,4 +34,8 @@
             + "Identifies which version of the Terms and Conditions the applicant accepted.")
     private String webUIVersion;
 
+    @JsonProperty("eligibilityOverrideOutcome")
+    @ApiModelProperty(notes = "Overrides the eligibility outcome from DWP or HMRC")
```
dwybourn: Should we just say the eligibility service?

YetAnotherMatt: Yes, we'll address that in the next PR

PR #533, opened 2020-01-13
Feature/AFHS-565 get claim by id endpoint

adeel718:


api/build.gradle
```diff
@@ -126,6 +126,9 @@ pmdTest {
     ruleSetFiles = files("../config/pmd/test-rulesets.xml")
     ruleSets = [] // Remove built-in rule sets
 }
+spotbugsMain {
+    excludeFilter = file("../config/spotbugs/main-exclude.xml")
```
mikeSmith08: What does this do?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceAssertionUtils.java
```diff
@@ -38,12 +37,18 @@ public static void assertClaimantMatchesClaimantDTO(ClaimantDTO claimant, Claima
...
-    public static RequestEntity buildClaimRequestEntity(Object requestObject) {
+    public static RequestEntity buildCreateClaimRequestEntity(Object requestObject) {
         var headers = new HttpHeaders();
         headers.setContentType(MediaType.APPLICATION_JSON);
         return new RequestEntity<>(requestObject, headers, HttpMethod.POST, CLAIMANT_ENDPOINT_URI_V3);
     }
 
+    public static RequestEntity buildRetrieveClaimRequestEntity(UUID claimId) {
+        var headers = new HttpHeaders();
+        headers.setContentType(MediaType.APPLICATION_JSON);
```
mikeSmith08: Can we refactor the header creation into a method which can be reused in these two methods, it can then be inlined in the return statement which makes these much more concise and removes the duplication.

api/build.gradle
YetAnotherMatt: We haven't had to configure this for any other project - are you sure it's necessary?

YetAnotherMatt: According to the spotbugs gradle plugin docs: "Current version of SpotBugs Gradle Plugin uses the same way with FindBugs Gradle Plugin to configure. Please check the document for FindBugsExtension." The findbugs extension docs provide these default values:
```
    findbugs {
        toolVersion = "2.0.1"
        sourceSets = [sourceSets.main]
        ignoreFailures = true
        reportsDir = file("$project.buildDir/findbugsReports")
        effort = "max"
        reportLevel = "high"
        visitors = ["FindSqlInjection", "SwitchFallthrough"]
        omitVisitors = ["FindNonShortCircuit"]
        includeFilter = file("$rootProject.projectDir/config/findbugs/includeFilter.xml")
        excludeFilter = file("$rootProject.projectDir/config/findbugs/excludeFilter.xml")
    }
```
So if it's not working without configuring the excludeFilter then we may need to rename the 'spotbugs' directory to be 'findbugs'.

YetAnotherMatt: actually, I just realised that our filenames are different - using different files for main and test. Presumably that means none of our spotbugs exclusions were working until now?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimDTO.java
```diff
@@ -48,7 +48,7 @@
     @JsonProperty("hmrcHouseholdIdentifier")
     private String hmrcHouseholdIdentifier;
 
-    @ApiModelProperty(notes = "Card id return from the card provider", example = "8b66a86f-778e-47f9-b998-a042476ef2c0 \n")
+    @ApiModelProperty(notes = "Card id returned from the card provider", example = "8b66a86f-778e-47f9-b998-a042476ef2c0 \n")
```
YetAnotherMatt: Probably shouldn't have a newline character at the end of the notes.

api/build.gradle
adeel718: We had to add this in because the current spotBugsTest task only excludes for tests, whereas we need to ignore a spotbug check when running the actual build. 
Because I have added a log message 'log.debug("Retrieve claim by id {}", id);' in the read API, id is passed in as a path variable, spotbug returns an error on this as this can cause crlf injection in the logs (even though this is a UUID). The option were to:
- Exclude this code from spotbug checks
- Add some regex to the id to replace and crlf which make it bit messy
- Add pattern to logging so that it strips crlf
- Annotate with @Suppressfbwarnings

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/ClaimDTO.java
adeel718: Done

PR #534, opened 2020-01-13
Feature/embedded database provider

YetAnotherMatt:
Embedded data source wasn't working in Mac laptop (Amar's) to run all tests. It works with docker provider - however it is defaulted to zonky to ensure it works in the CI build


mikeSmith08:
Does this mean that Docker is required to run the tests that require an embedded database now?

YetAnotherMatt:
> Does this mean that Docker is required to run the tests that require an embedded database now?

No - only if you set the environment variable

PR #538, opened 2020-01-14
AFHS-405 Split payment processing to request and complete payments

adeel718:
To split the make payment processor to
- Request payment balance and store referenceId
- Complete payment and making the actual payment to provider with stored referenceId


api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/MessageType.java
```diff
@@ -9,5 +9,7 @@
     DETERMINE_ENTITLEMENT,
     ADDITIONAL_PREGNANCY_PAYMENT,
     REPORT_CLAIM,
-    REPORT_PAYMENT
+    REPORT_PAYMENT,
+    REQUEST_PAYMENT,
```
mikeSmith08: Presumably the last part of this will be to remove the `MAKE_PAYMENT` enum and processor?

adeel718: Correct, but I couldn't do it as it would break everything. So removing MAKE_PAYMENT will be a different pr

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/CompletePaymentMessageContext.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Value;
+import uk.gov.dhsc.htbhf.claimant.entity.Payment;
+import uk.gov.dhsc.htbhf.claimant.entity.PaymentCycle;
+
+@Value
+@Builder
+public class CompletePaymentMessageContext {
+    private PaymentCycle paymentCycle;
+    private Payment payment;
+    private String referenceId;
```
mikeSmith08: I think this `referenceId` should have a more specific name as we have quite a lot of reference ids in the the system, how about `paymentProviderReferenceId` or just `paymentReferenceId`?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
```diff
@@ -20,6 +20,7 @@
 @Slf4j
 @RequiredArgsConstructor
 @Component
+@SuppressWarnings("PMD.TooManyMethods")
```
mikeSmith08: I think this is fine for a class like this, splitting this into multiple components just to keep PMD happy isn't the right thing to do.

mikeSmith08:
This is an odd way to split up the work in the PRs, I'd have been tempted to just do one processor first and not plug it in at all so we can verify if the payload and context objects are correct, then do the second processor in the next PR and then plug in with the schedules last. Just my thoughts.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/CompletePaymentMessageContext.java
YetAnotherMatt: `paymentReferenceId` gets my vote

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/CompletePaymentMessageProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Component
+@AllArgsConstructor
+@Slf4j
+public class CompletePaymentMessageProcessor implements MessageTypeProcessor {
+
+    @Override
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public MessageStatus processMessage(Message message) {
+
```
YetAnotherMatt: shouldn't this be processing the message?

YetAnotherMatt: Or have a TODO statement, at least

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/CompletePaymentMessageContext.java
dwybourn: How about depositFundsReferenceId

api/src/main/java/uk/gov/dhsc/htbhf/claimant/scheduler/MessageProcessorScheduler.java
YetAnotherMatt: If the methods contained logic then we probably would want to split it up. As they're just scheduling tasks it makes sense to me to keep them together.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/context/CompletePaymentMessageContext.java
adeel718: Changed to paymentReferenceId

api/src/main/java/uk/gov/dhsc/htbhf/claimant/message/processor/CompletePaymentMessageProcessor.java
adeel718: Added TODO's

PR #543, opened 2020-01-16
Added EligibilityOverride object with EligibilityOverrideOutcome property

AmarkumarJagtap:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claim.java
```diff
@@ -96,6 +90,9 @@
     @Valid
     private Claimant claimant;
 
+    @Embedded
```
dwybourn: Why has this been made an embedded object?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
```diff
@@ -52,13 +51,8 @@
     @ApiResponses({@ApiResponse(code = 400, message = "Bad request", response = ErrorResponse.class)})
     public ResponseEntity<ClaimResultDTO> createClaim(@RequestBody @Valid @ApiParam("The claim to persist") NewClaimDTO newClaimDTO) {
         log.debug("Received claim");
-        Claimant claimant = claimantConverter.convert(newClaimDTO.getClaimant());
```
dwybourn: Whilst I agree with this change, it seems unrelated to this story and in future should be done in it's own PR.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/converter/NewClaimDTOToClaimRequestConverter.java
```diff
@@ -0,0 +1,34 @@
...
+        return ClaimRequest.builder()
+                .claimant(claimant)
+                .deviceFingerprint(dto.getDeviceFingerprint())
+                .webUIVersion(dto.getWebUIVersion())
+                .eligibilityOverride(eligibilityOverride)
+                .build();
+    }
+
+    private EligibilityOverride convertToEligibilityOverride(EligibilityOverrideDTO dto) {
+        return dto == null ? null : EligibilityOverride.builder()
```
dwybourn: Can we update the formatting here please, the builder should really be on it's own line.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/converter/NewClaimDTOToClaimRequestConverterTest.java
```diff
@@ -0,0 +1,56 @@
...
+        assertThat(claimRequest).isNotNull();
+        assertThat(claimRequest.getClaimant()).isEqualTo(claimant);
+        assertThat(claimRequest.getDeviceFingerprint()).isEqualTo(dto.getDeviceFingerprint());
+        assertThat(claimRequest.getWebUIVersion()).isEqualTo(dto.getWebUIVersion());
+        assertEligibilityOverride(claimRequest.getEligibilityOverride(), dto.getEligibilityOverride());
+
+    }
+
+    private void assertEligibilityOverride(EligibilityOverride eligibilityOverride, EligibilityOverrideDTO dto) {
+        if (null == dto) {
```
dwybourn: We shouldn't have a check like this in the test as we know whether or not the EligibilityOverrideDTO is null or not. 
We should have one test for null, and one for not null.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/claim/ClaimServiceTest.java
```diff
@@ -42,32 +43,51 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static uk.gov.dhsc.htbhf.TestConstants.*;
-import static uk.gov.dhsc.htbhf.claimant.message.payload.LetterType.*;
+import static uk.gov.dhsc.htbhf.claimant.message.payload.LetterType.APPLICATION_SUCCESS_CHILDREN_MATCH;
+import static uk.gov.dhsc.htbhf.claimant.message.payload.LetterType.APPLICATION_SUCCESS_CHILDREN_MISMATCH;
+import static uk.gov.dhsc.htbhf.claimant.message.payload.LetterType.UPDATE_YOUR_ADDRESS;
 import static uk.gov.dhsc.htbhf.claimant.model.eligibility.EligibilityAndEntitlementDecision.buildDuplicateDecisionWithExistingClaimId;
-import static uk.gov.dhsc.htbhf.claimant.testsupport.ClaimRequestTestDataFactory.*;
```
dwybourn: Can you update your intellij static import settings please so we don't keep getting these diffs in the PRs.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/converter/NewClaimDTOToClaimRequestConverter.java
mikeSmith08: Yes, I agree. As an alternative, I'd prefer it to look something like this:

```
    private EligibilityOverride convertToEligibilityOverride(EligibilityOverrideDTO dto) {
        if(dto == null) {
            return null;
        }
        return EligibilityOverride.builder()
                .eligibilityOutcome(dto.getEligibilityOutcome())
                .build();
    }
```
This then very clear what is going on and the alternate flow returns quickly.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.dhsc.htbhf.claimant.testsupport;
+
+import uk.gov.dhsc.htbhf.claimant.entity.EligibilityOverride;
+import uk.gov.dhsc.htbhf.dwp.model.EligibilityOutcome;
+
+public class EligibilityOverrideTestDataFactory {
+
+    public static EligibilityOverride eligibilityConfirmedWithNoChildrenForFiveYears() {
```
mikeSmith08: I don't see what FiveYears has to do with this object? Making a guess, it could either be how long the override applies for or the age of children, but its unclear. Either way, has nothing to do with the method, so I'd rename it to something simpler like `aConfirmedEligibilityOverride`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
```diff
@@ -0,0 +1,19 @@
...
+        return EligibilityOverride.builder()
+                .eligibilityOutcome(EligibilityOutcome.CONFIRMED)
+                .build();
+    }
+
+    public static EligibilityOverride eligibilityNotConfirmed() {
```
mikeSmith08: I think this would be better as `aNotConfirmedEligibilityOverride`, this method name sounds more like an EligibilityOutcome that is being created.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
```diff
@@ -155,19 +149,7 @@ void shouldReturnInternalServerErrorStatusWhenEligibilityStatusIsError() {
         assertThat(response).isNotNull();
         assertThat(response.getStatusCode()).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR);
         assertThat(response.getBody()).isEqualTo(aClaimResultDTOWithClaimStatusAndNoVoucherEntitlement(ClaimStatus.ERROR));
-        verifyCreateClaimCalledCorrectly(claimant, dto);
-        verify(claimantConverter).convert(aValidClaimantDTO());
-    }
-
-    private void verifyCreateClaimCalledCorrectly(Claimant claimant, NewClaimDTO dto) {
```
mikeSmith08: I'm not sure why you removed this method, it seems like a perfectly valid thing to do to make sure that the ClaimService was called correctly, I think this should be added back into these tests.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -77,26 +80,24 @@
...
     }
 
-    private static Stream<EligibilityOutcome> allEligibilityOutcomeValues() {
+    private static Stream<EligibilityOverride> allEligibilityOutcomeValues() {
         return Stream.of(
                 null,
-                EligibilityOutcome.CONFIRMED,
-                EligibilityOutcome.NOT_CONFIRMED,
-                EligibilityOutcome.NOT_SET
+                eligibilityConfirmedWithNoChildrenForFiveYears()
```
mikeSmith08: Shouldn't we add in values for the other EligibilityOutcomes within EligibilityOverride objects, seems like we've deleted some tests here by doing this.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/ClaimRequestTestDataFactory.java
```diff
@@ -18,8 +18,8 @@ public static ClaimRequest aValidClaimRequest() {
         return aClaimRequestBuilderForClaimant(aClaimantWithChildrenDob(MAGGIE_AND_LISA_DOBS)).build();
     }
 
-    public static ClaimRequest aClaimRequestWithEligibilityOverrideOutcome(Claimant claimant, EligibilityOutcome eligibilityOverrideOutcome) {
-        return aClaimRequestBuilderForClaimant(claimant).eligibilityOverrideOutcome(eligibilityOverrideOutcome).build();
+    public static ClaimRequest aClaimRequestWithEligibilityOverrideOutcome(Claimant claimant, EligibilityOverride eligibilityOverride) {
```
mikeSmith08: This should be renamed to `aClaimRequestWithEligibilityOverride`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/NewClaimDTOTestDataFactory.java
```diff
@@ -51,12 +46,16 @@ public static NewClaimDTO aValidClaimDTOWithNoNullFields() {
                 .build();
     }
 
-    public static NewClaimDTO aValidClaimDTOWithEligibilityOverrideOutcome(LocalDate expectedDeliveryDate,
-                                                                           List<LocalDate> childrenDob,
-                                                                           EligibilityOutcome eligibilityOutcome) {
+    public static NewClaimDTO aValidClaimDTOWithEligibilityOverride(LocalDate expectedDeliveryDate,
```
mikeSmith08: This should be `aValidClaimDTOWithEligibilityOutcome` as that is all you are passing in as a parameter.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/NewClaimDTOTestDataFactory.java
```diff
@@ -51,12 +46,16 @@ public static NewClaimDTO aValidClaimDTOWithNoNullFields() {
...
+                                                                    List<LocalDate> childrenDob,
+                                                                    EligibilityOutcome eligibilityOutcome) {
+        EligibilityOverrideDTO  eligibilityOverride = EligibilityOverrideDTO.builder()
```
mikeSmith08: This might need a test data factory.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/EligibilityOverrideDTO.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.dhsc.htbhf.dwp.model.EligibilityOutcome;
+
+import javax.validation.constraints.NotNull;
+
+@Data
+@Builder
+@AllArgsConstructor(onConstructor_ = {@JsonCreator})
+@ApiModel(description = "Eligibility override decision based on eligibility outcome and until date ")
```
mikeSmith08: There isn't a date yet on this object.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/entity/Claim.java
YetAnotherMatt: As there will be 3 properties (eligiilityOutcome, overrideUntil and datesOfBirth) that shouls all be present or all absent, an object to encapsulate them felt right.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
YetAnotherMatt: Fair point. We did it because we were at the point of creating a new converted for the EligibilityOverride, and this felt cleaner. Should have been in a separate pr though.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/claim/ClaimServiceTest.java
YetAnotherMatt: The import settings are now correct (5/5) - this is fixing old commits.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
YetAnotherMatt: We're no longer constructing a Claimrequest object in the controller, and this method was testing that the ClaimRequest was constructed correctly.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/converter/NewClaimDTOToClaimRequestConverter.java
AmarkumarJagtap: yes, agreed

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
mikeSmith08: Let me rephrase that then, you are no longer verifying that the claim service is being called by not using this method, so that should be reinstated.

mikeSmith08: You've missed my point, we are no longer verifying that the ClaimService is being called with a `verify()` call on that mock, we should be verifying that.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
YetAnotherMatt: I think we need to rename the method. We're actually only testing that a duplicate outcome is returned with or without an eligibilityOverride

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
AmarkumarJagtap: added verify on create claim

api/src/test/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimControllerTest.java
```diff
@@ -155,19 +152,8 @@ void shouldReturnInternalServerErrorStatusWhenEligibilityStatusIsError() {
         assertThat(response).isNotNull();
         assertThat(response.getStatusCode()).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR);
         assertThat(response.getBody()).isEqualTo(aClaimResultDTOWithClaimStatusAndNoVoucherEntitlement(ClaimStatus.ERROR));
-        verifyCreateClaimCalledCorrectly(claimant, dto);
```
mikeSmith08: You need to verify the claim service is called here too.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -75,31 +75,28 @@
...
+    void shouldReturnDuplicateWhenLiveClaimAlreadyExistsWithoutEligibilityOverride() {
+        shouldReturnDuplicateWhenLiveClaimAlreadyExists(null);
+    }
+
+    @Test
+    void shouldReturnDuplicateWhenLiveClaimAlreadyExistsWithEligibilityOverride() {
+        aConfirmedEligibilityOverrideWithNoChildren();
+    }
+
+    private void shouldReturnDuplicateWhenLiveClaimAlreadyExists(EligibilityOverride eligibilityOverride) {
```
mikeSmith08: This is only used in one test, so this would be better inlined, especially as the only value used for EligibilityOverride is null.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -75,31 +75,28 @@
...
-    @MethodSource("allEligibilityOutcomeValues")
-    void shouldReturnDuplicateWhenLiveClaimAlreadyExists(EligibilityOutcome eligibilityOverrideOutcome) {
+    @Test
+    void shouldReturnDuplicateWhenLiveClaimAlreadyExistsWithoutEligibilityOverride() {
+        shouldReturnDuplicateWhenLiveClaimAlreadyExists(null);
+    }
+
+    @Test
+    void shouldReturnDuplicateWhenLiveClaimAlreadyExistsWithEligibilityOverride() {
+        aConfirmedEligibilityOverrideWithNoChildren();
```
mikeSmith08: This does nothing, what has happened to this test?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.dhsc.htbhf.claimant.testsupport;
+
+import uk.gov.dhsc.htbhf.claimant.entity.EligibilityOverride;
+import uk.gov.dhsc.htbhf.dwp.model.EligibilityOutcome;
+
+public class EligibilityOverrideTestDataFactory {
+
+    public static EligibilityOverride aConfirmedEligibilityOverrideWithNoChildren() {
```
mikeSmith08: This EligibilityOverride has nothing to do with children so should be renamed to just `aConfirmedEligibilityOverride`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
AmarkumarJagtap: Missed to add shouldReturnDuplicateWhenLiveClaimAlreadyExists with eligibility override not null.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
AmarkumarJagtap: One test with null and other with valid eligibility override object, Comment above should resolve this

PR #549, opened 2020-01-20
AFHS-429 - added children dob list in eligibility override

AmarkumarJagtap:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/entity/constraint/EligibilityOverrideValidatorTest.java
```diff
@@ -47,14 +48,22 @@ void shouldFailToValidateEligibilityOverrideWithNullValues(EligibilityOverride e
                 EligibilityOverride.builder()
                         .eligibilityOutcome(null)
                         .overrideUntil(OVERRIDE_UNTIL_FIVE_YEARS)
+                        .childrenDob(emptyList())
```
mikeSmith08: I think this would be clearer with specific dates like `MAGGIE_AND_LISA_DOBS` rather than `emptyList()`, same below too.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/NewClaimDTOTest.java
```diff
@@ -111,13 +130,16 @@ void shouldFailToValidateClaimWithEligibilityOverrideUntilDatesNotInFuture(Local
     @Test
     void shouldSuccessfullyValidateClaimWithEligibilityOverride() {
         //Given
-        NewClaimDTO claim = aValidClaimDTOWithEligibilityOverride(
-                EXPECTED_DELIVERY_DATE_IN_TWO_MONTHS,
-                NO_CHILDREN,
-                EligibilityOutcome.CONFIRMED,
-                OVERRIDE_UNTIL_FIVE_YEARS);
+        EligibilityOverrideDTO eligibilityOverrideDTO = EligibilityOverrideDTO.builder()
```
mikeSmith08: Can we have this in a test data factory please?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -174,10 +179,12 @@ void shouldEvaluateClaimForGivenPaymentCycle() {
         verify(eligibilityAndEntitlementDecisionFactory).buildDecision(IDENTITY_AND_ELIGIBILITY_RESPONSE, VOUCHER_ENTITLEMENT, false);
     }
 
-    @Test
-    void shouldEvaluateClaimForGivenPaymentCycleWithEligibilityOverride() {
+    @ParameterizedTest
+    @MethodSource("childrenDobs")
```
mikeSmith08: We expect no children to behave the same as having children looking at this test, just wondering why the need to parameterise this test?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
```diff
@@ -4,27 +4,39 @@
...
 
 import java.time.LocalDate;
+import java.util.List;
 
+import static java.util.Collections.emptyList;
 import static uk.gov.dhsc.htbhf.claimant.testsupport.TestConstants.OVERRIDE_UNTIL_FIVE_YEARS;
 
 public class EligibilityOverrideTestDataFactory {
 
     public static EligibilityOverride aConfirmedEligibilityOverride() {
```
mikeSmith08: How about `aConfirmedEligibilityOverrideWithNoChildren`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
```diff
@@ -4,27 +4,39 @@
...
+    }
+
+    public static EligibilityOverride aConfirmedEligibilityOverrideWithChildren(List<LocalDate> childrenDob) {
+        return aConfirmedEligibilityOverrideBuilder()
+                .overrideUntil(OVERRIDE_UNTIL_FIVE_YEARS)
+                .childrenDob(childrenDob)
                 .build();
     }
 
     public static EligibilityOverride aConfirmedEligibilityWithUntilDate(LocalDate overrideUntil) {
```
mikeSmith08: Would it be worthwhile mentioning no children in this and the next method too?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -26,5 +29,6 @@
     public static final String TEST_EXCEPTION_MESSAGE = "test exception";
     public static final RuntimeException TEST_EXCEPTION = new RuntimeException(TEST_EXCEPTION_MESSAGE);
     public static final LocalDate OVERRIDE_UNTIL_FIVE_YEARS = LocalDate.now().plusYears(5);
+    public static final List<LocalDate> CHILD_BORN_IN_FUTURE = singletonList(LocalDate.now().plusDays(1));
```
mikeSmith08: Could we call this `CHILD_DOB_IN_FUTURE` please as being born is the future is somewhat impossible. :-)

swagger.yml
```diff
@@ -378,6 +378,12 @@ definitions:
   EligibilityOverrideDTO:
     type: "object"
     properties:
+      childrenDob:
+        type: "array"
+        description: "The dates of birth of the claimant's children (if they have\
```
dwybourn: Can we update the description to state that an empty list must be used if there are no children.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
YetAnotherMatt: It's only used to validate that the future DOB is rejected - so I think it being impossible is valid :)

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
YetAnotherMatt: we pass the list of childrenDobs into a constructor _and an assertion_ - so each test does assert that different children's dobs are used. Without adding this the original test would have passed if the code had hard-coded an empty list. I guess we could simply use a non-empty list, but the parameterised test doesn't hurt.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/EligibilityOverrideTestDataFactory.java
YetAnotherMatt: Not for `aNotConfirmedEligibilityOverride` - there should never be children if eligibility is not confirmed.

PR #550, opened 2020-01-20
AFGS-405 Added payment factory.

dwybourn:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/payments/PaymentServiceTest.java
```diff
@@ -184,21 +184,16 @@ private PaymentCycle createAndSavePaymentCycle() {
     }
 
     private void verifyFailedPaymentSavedWithAllData(PaymentCycle paymentCycle, int amountToPay, String paymentRequestReference, FailureEvent failureEvent) {
-        Payment actualPayment = verifyFailedPaymentSavedCorrectly(paymentCycle, failureEvent);
```
dwybourn: github diff isn't very clear here. The payment assertions was unnecessarily being split into two methods so they've been collapsed into one. 
A count of the total payments is now asserted.
No longer checking the id is not null as we already fetched it via the paymentRepository.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/FailureEventTestDataFactory.java
```diff
@@ -14,4 +16,13 @@ public static FailureEvent aFailureEventWithEvent(Event event) {
                 .exception(TEST_EXCEPTION)
                 .build();
     }
+
+    public static FailureEvent aFailedMakePaymentEvent(PaymentCycle paymentCycle, int amountToPay, String requestReference) {
```
adeel718: Is there going to be audits for RequestPayment and CompletePayment since MakePayment has now been removed?

api/src/main/java/uk/gov/dhsc/htbhf/claimant/factory/PaymentFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Creates a new payment with a status of {@link PaymentStatus#SUCCESS}. To be used when making a payment was successful.
+     * @param paymentCycle Payment cycle linked to the payment
+     * @param paymentCalculation payment calculation used when making the payment
+     * @param paymentResult the result of the payment
+     * @return new payment
+     */
+    public static Payment createCompletedPayment(PaymentCycle paymentCycle, PaymentCalculation paymentCalculation, PaymentResult paymentResult) {
```
mikeSmith08: Should this be `createSuccessfulPayment` then as technically a failed payment is completed too, it just has a different status.

dwybourn: That's true, renamed to createSuccessfulPayment

PR #552, opened 2020-01-21
AFGS-405 Added payment cycle integration tests that send different pa…

dwybourn:
…yment amounts depending on the claimant's balance.


api/src/test/java/uk/gov/dhsc/htbhf/claimant/ScheduledServiceIntegrationTest.java
```diff
@@ -124,6 +125,11 @@ void assertThatRegularPaymentEmailWasSent(PaymentCycle newCycle) throws Notifica
         assertThatPaymentEmailWasSent(newCycle, EmailType.REGULAR_PAYMENT);
     }
 
+    void assertThatRegularPaymentEmailWasSentOnly(PaymentCycle paymentCycle) throws NotificationClientException {
```
YetAnotherMatt: Can I suggest that we move `assertSingleEmailSent` from ClaimantLifecycleIntegrationTests into this class, and use it instead of this method?

dwybourn: assertThatRegularPaymentEmailWasSent() is specific to payment emails and performs additional assertions to make sure the email personalisation is correct for a payment. assertSingleEmailSent doesn't have any of these payment checks.

PR #555, opened 2020-01-22
AFHS-748- updated Claimant to be embeddable in claim 

AmarkumarJagtap:
and created sql script to add claimant columns in claim


api/src/test/java/uk/gov/dhsc/htbhf/claimant/model/UpdatableClaimantFieldTest.java
```diff
@@ -44,7 +44,6 @@ void shouldReportFieldsDoNotHaveDifferentValues(UpdatableClaimantField field) {
     void shouldReportNullFieldsDoNotHaveDifferentValues(UpdatableClaimantField field) {
         Claimant originalClaimant = Claimant.builder().build();
         Claimant newClaimant = Claimant.builder().build();
-        assertThat(newClaimant).isNotEqualTo(originalClaimant);
 
         assertThat(field.valueIsDifferent(originalClaimant, newClaimant)).isFalse();
```
mikeSmith08: Shouldn't we now be doing `assertThat(newClaimant).isEqualTo(originalClaimant);` I'm not sure that this assertion is any more helpful (ore required) than just using `isEqualTo`.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/RepositoryMediator.java
```diff
@@ -149,7 +149,6 @@ public void ageDatabaseEntities(int numberOfDays) {
         paymentRepository.findAll().forEach(payment -> EntityAgeAccelerator.ageObject(payment, numberOfDays));
         claimRepository.findAll().forEach(claim -> {
             EntityAgeAccelerator.ageObject(claim, numberOfDays);
-            EntityAgeAccelerator.ageObject(claim.getClaimant(), numberOfDays);
```
mikeSmith08: Why do we now no longer need to age the Claimant, did we even need to do it in the first place?

YetAnotherMatt: We did age the claimant before, but it's now being aged via the claim. We removed this line to prevent claimants being aged twice. The EntityAgeAccelerator ages Temporal fields in the object, and all child (non-entity) objects. Claim is no longer an entity, so it is included when we age the claim.

PR #556, opened 2020-01-22
sql script to remove claimant_id from claim and drop claimant table

AmarkumarJagtap:


db/src/main/resources/db/migration/V1_058__delete_claimant_table.sql
```diff
@@ -0,0 +1,15 @@
+update claim  set
```
YetAnotherMatt: Running this query again to catch any claims created by old versions of the service during blue/green deployment

PR #558, opened 2020-01-22
AFHS-1762 - Null pointer exception bug in ReportPropertiesFactory 

AmarkumarJagtap:
When reporting a claim and creating the common custom dimensions for calling google analytics, we obtain the children's DOB from the identityAndEligibilityresponse. However, if a claim was rejected due to a duplicate NINO, we never call the eligibility service, so this object is never populated. The factory should instead check for a null value and obtain the children's dob from the claim instead.

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
@@ -119,10 +119,11 @@ public abstract class ReportPropertiesFactory {
 
     protected Map<String, Object> createCommonCustomMetrics(ReportEventMessageContext context) {
         Map<String, Object> customMetrics = new TreeMap<>();
-        Claimant claimant = context.getClaim().getClaimant();
+        Claim claim = context.getClaim();
+        Claimant claimant = claim.getClaimant();
         LocalDate atDate = context.getTimestamp().toLocalDate();
         customMetrics.put(CLAIMANT_AGE.getFieldName(), Period.between(claimant.getDateOfBirth(), atDate).getYears());
-        List<LocalDate> dobOfChildrenUnder4 = getDobOfChildrenUnder4(context.getClaim(), context.getIdentityAndEligibilityResponse());
+        List<LocalDate> dobOfChildrenUnder4 = getDobOfChildrenUnder4(claim, context.getIdentityAndEligibilityResponse());
         long childrenUnder4 = getNumberOfChildrenUnderFour(dobOfChildrenUnder4, atDate);
         long childrenUnder1 = getNumberOfChildrenUnderOne(dobOfChildrenUnder4, atDate);
         customMetrics.put(CHILDREN_UNDER_ONE.getFieldName(), childrenUnder1);
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
@@ -5,6 +5,8 @@ import uk.gov.dhsc.htbhf.claimant.entity.Claimant;
 import uk.gov.dhsc.htbhf.claimant.message.context.ReportEventMessageContext;
 import uk.gov.dhsc.htbhf.claimant.model.PostcodeData;
 import uk.gov.dhsc.htbhf.claimant.reporting.ClaimantCategoryCalculator;
+import uk.gov.dhsc.htbhf.dwp.model.QualifyingBenefits;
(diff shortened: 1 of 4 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
```diff
@@ -95,17 +100,29 @@ protected Long calculateQueueTime(LocalDateTime timestamp) {
...
+                : identityAndEligibilityResponse.getQualifyingBenefits();
+    }
+
     protected Map<String, Object> createCommonCustomMetrics(ReportEventMessageContext context) {
         Map<String, Object> customMetrics = new TreeMap<>();
         Claimant claimant = context.getClaim().getClaimant();
         LocalDate atDate = context.getTimestamp().toLocalDate();
         customMetrics.put(CLAIMANT_AGE.getFieldName(), Period.between(claimant.getDateOfBirth(), atDate).getYears());
-        List<LocalDate> dobOfChildrenUnder4 = context.getIdentityAndEligibilityResponse().getDobOfChildrenUnder4();
+        List<LocalDate> dobOfChildrenUnder4 = getDobOfChildrenUnder4(context.getClaim(), context.getIdentityAndEligibilityResponse());
```
adeel718: Can context.getClaim be declared as a variable instead as it's now used more than once

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
```diff
@@ -95,17 +100,29 @@ protected Long calculateQueueTime(LocalDateTime timestamp) {
...
-        customDimensions.put(QUALIFYING_BENEFIT.getFieldName(), context.getIdentityAndEligibilityResponse().getQualifyingBenefits());
+        customDimensions.put(QUALIFYING_BENEFIT.getFieldName(), getQualifyingBenefits(identityAndEligibilityResponse));
 
         return customDimensions;
     }
 
+    private List<LocalDate> getDobOfChildrenUnder4(Claim claim, CombinedIdentityAndEligibilityResponse identityAndEligibilityResponse) {
+        return identityAndEligibilityResponse == null
+                ? claim.getClaimant().getInitiallyDeclaredChildrenDob()
+                : identityAndEligibilityResponse.getDobOfChildrenUnder4();
```
mikeSmith08: This is a change in behaviour here rather than fixing a bug, this is saying that if there aren't any children returned from the eligibility service then we trust what the Claimant has told us, is that what you intended? If so, this feels like it should be in its own story and documented in business rules somewhere rather than fixing a NPE bug.

dwybourn: That's not quite correct, it's saying if there's no identityAndEligibilityResponse (due to not calling the eligibility service when the claim's a duplicate), we get the initially declared children. Before we just threw a NPE and retried the message.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportClaimPropertiesFactoryTest.java
```diff
@@ -197,4 +217,14 @@ private ReportClaimMessageContext aReportClaimMessageContext(LocalDateTime times
                 .identityAndEligibilityResponse(anIdMatchedEligibilityConfirmedUCResponseWithAllMatches(datesOfBirthOfChildren))
                 .timestamp(timestamp);
     }
+
+    private ReportClaimMessageContext aReportClaimMessageContextWithoutDecision(LocalDateTime timestamp, List<LocalDate> datesOfBirthOfChildren) {
```
mikeSmith08: Should this be in a test data factory.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/reporting/payload/ReportPropertiesFactory.java
mikeSmith08: Yes, I can see that, that's what I was trying to say. Still seems like way more than fixing a NPE, if that's what was being done then it should just handle the null and return null but not change the logic to use the initially declared children.

dwybourn: @YetAnotherMatt This is something I wanted to mention before you left. With the PR as is, you can't tell them difference in GA between a claim with children dobs from DWP or from the claimant.

YetAnotherMatt: Given that it's only used for reporting - and if the identityAndEligibilityResponse is null we're reporting a rejected claim - I don't see it as an issue.

PR #559, opened 2020-01-23
AFHS-564 Created partial unique index for nino on claim table and updated integration tests

AmarkumarJagtap:


api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
```diff
@@ -241,6 +249,36 @@ void shouldReturnDuplicateStatusWhenEligibleClaimAlreadyExistsForDwpHouseholdIde
         wiremockManager.assertThatEligibilityRequestMade();
     }
 
+    @Test
+    void shouldReturnDuplicateStatusWhenTwoEligibleClaimsForSameNino() throws JsonProcessingException, InterruptedException {
```
YetAnotherMatt: Minor, but suggest a slight change in wording: `shouldReturnDuplicateStatusWhenTwoSimultaneousClaimsForSameNino`

db/src/main/resources/db/migration/V1_059__partial_unique_index_nino_claim_table.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX unique_nino_for_active_claim ON claim (nino) WHERE (claim_status in ('NEW', 'ACTIVE', 'PENDING', 'PENDING_EXPIRY'));
```
YetAnotherMatt: Minor, but unnecessary brackets around `claim_status in ('NEW', 'ACTIVE', 'PENDING', 'PENDING_EXPIRY')`

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
```diff
@@ -241,6 +249,36 @@ void shouldReturnDuplicateStatusWhenEligibleClaimAlreadyExistsForDwpHouseholdIde
...
+                () -> restTemplate.exchange(buildCreateClaimRequestEntity(dto), ClaimResultDTO.class),
+                () -> restTemplate.exchange(buildCreateClaimRequestEntity(dto), ClaimResultDTO.class)));
+
+        List<ResponseEntity<ClaimResultDTO>> results = futures.stream().map(objectFuture -> {
+            try {
+                return (ResponseEntity<ClaimResultDTO>) objectFuture.get();
+            } catch (InterruptedException | ExecutionException e) {
+                throw new RuntimeException(e);
+            }
+        }).collect(Collectors.toList());
```
mikeSmith08: Using `CompletableFutures` here would make this a lot tidier, can I suggest trying to use those instead?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
```diff
@@ -241,6 +249,36 @@ void shouldReturnDuplicateStatusWhenEligibleClaimAlreadyExistsForDwpHouseholdIde
...
+
+        //Then
+        //one transaction will succeed because it commits first.
+        assertThat(results.stream().anyMatch(r -> r.getBody().getClaimStatus() == ClaimStatus.NEW)).isTrue();
```
mikeSmith08: This assertion will also pass if both statuses are NEW, you need to make sure that you have one with NEW and one with REJECTED.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
```diff
@@ -241,6 +249,36 @@ void shouldReturnDuplicateStatusWhenEligibleClaimAlreadyExistsForDwpHouseholdIde
...
+
+        //one transaction will fail causing error because there is already an active claim with same nino.
+        assertThat(results.stream().anyMatch(r -> r.getStatusCode() == HttpStatus.INTERNAL_SERVER_ERROR)).isTrue();
```
mikeSmith08: Similarly here you need to specifically check that you have INTERNAL_SERVER_ERROR and OK, this only checks that at least one status code is INTERNAL_SERVER_ERROR.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
AmarkumarJagtap: There is one more assert which ensures that one matches with New and other with 500

dwybourn: Whilst it shouldn't be possible with our code, the two anyMatch could be asserting on the same object as they assert on separate fields. Really they should both be asserting on claimStatus and statusCode.

mikeSmith08: You are mixing up ClaimStatus and HttpStatsCode here, you need to have two assertions for each status, you are currently doing one assertion for each status.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -31,4 +31,7 @@
     public static final LocalDate OVERRIDE_UNTIL_FIVE_YEARS = LocalDate.now().plusYears(5);
     public static final List<LocalDate> CHILD_BORN_IN_FUTURE = singletonList(LocalDate.now().plusDays(1));
 
+    //NINOs
+    public static final String NINO = "MC125456C";
```
mikeSmith08: You do not need to do this, there are loads of test NINOs in the common-java repository, the idea was not to litter this file with them too, so lets remove this and use one of the standard ones (HOMER_NINO or MARGE_NINO for example).

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
YetAnotherMatt: how about if the first assertion is on a status code of 201 and claim status of NEW?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
YetAnotherMatt: We need a third NINO for one of the tests, I don't think we have a third currently?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantServiceIntegrationTests.java
YetAnotherMatt: Just to note that the other status will be CREATED, not OK.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
mikeSmith08: I'll add this one to common-java so we can remove it from here, leave that to me.

mikeSmith08: You do need to be careful with this NINO, it's the same as the HOME_NINO value so it may cause issues and failing test, so I'd suggest changing it.

mikeSmith08: Actually, I've just spotted it is very slightly different so its ok. I'll make it substantially different in the one in common-java so its more obvious.

PR #561, opened 2020-01-28
Feature/eligibility override json column

AmarkumarJagtap:
Qualifying benefits will be added to EligibilityOverride as part of AFHS-796 to set qualifying benefits to under_18.  
Converted existing 3 override columns to eligibility_override json column  to keep all 3 fields ( +1 in next pr) as part of json in eligibility_override. This is to avoid many columns in claim table  for eligibility override


mikeSmith08:
Is there any particular reason for making this change? A comment added to the PR usually helps prevent such questions.

AmarkumarJagtap:
> Is there any particular reason for making this change? A comment added to the PR usually helps prevent such questions.

added description now

PR #562, opened 2020-01-28
removed eligibility_override_outcome, eligibility_override_until, eligibility_override_children_dob columns from Claim table

AmarkumarJagtap:
Removed EligibilityOverrideValidator and updated test cases


db/src/main/resources/db/migration/V1_061__replace_eligiblity_override_columns_with_json.sql
```diff
@@ -0,0 +1,3 @@
+alter table claim drop column eligibility_override_outcome;
```
YetAnotherMatt: We should include the update sql again before we drop the columns

PR #563, opened 2020-01-29
added QualifyingReason to EligibilityOverride, DTO and converters

AmarkumarJagtap:


api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/EligibilityOverrideDTO.java
```diff
@@ -32,6 +33,10 @@
     @ApiModelProperty(notes = "The date at which the override expires", example = "2025-05-17")
     private LocalDate overrideUntil;
 
+    @JsonProperty("qualifyingBenefits")
+    @ApiModelProperty(notes = "Overrides the reason that this applicant qualifies for Healthy Start", example = "under_18")
```
mikeSmith08: I think this description really confuses matters because I would expect this to be their qualifying benefit not their reason for overriding. I understand why this has been done but it feels like a mistake to me, this enum now has two means and purposes and this API description highlights the confusion very well.

YetAnotherMatt: Would the confusion be resolved if we replaced `QualifyingBenefits` in the CombinedIdentityAndEligibilityResponse with a new `QualifyingReason` enum that has the same values as `QualifyingBenefit` plus `under_18`?

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
```diff
@@ -213,7 +236,7 @@ void shouldEvaluateClaimForGivenPaymentCycleWithEligibilityOverride(List<LocalDa
     @Test
     void shouldEvaluateClaimForGivenPaymentCycleWithExpiredEligibilityOverride() {
         //Given
-        Claim claim = aClaimWithEligibilityOverride(aConfirmedEligibilityWithNoChildrenOverriddenUntil(LocalDate.now()));
+        Claim claim = aClaimWithEligibilityOverride(aConfirmedEligibilityForPregnantTeenager(LocalDate.now()));
```
mikeSmith08: Why was this changed? Seems like the original test should stay intact.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/EligibilityOverrideDTO.java
mikeSmith08: Yes, that might work.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/service/EligibilityAndEntitlementServiceTest.java
AmarkumarJagtap: ah i think it got changed during refactor, i will revert this.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantLifecycleIntegrationTests.java
```diff
@@ -456,6 +494,17 @@ private UUID applyForHealthyStartOverridingEligibilityForAPregnantWomanWithNoChi
         return applyForHealthyStart(newClaimDTO, NO_CHILDREN);
     }
 
+    private UUID applyForHealthyStartOverridingEligibilityForAPregnantTeenager(LocalDate expectedDeliveryDate)
```
dwybourn: Can we rename the method to say under18 rather than Teenager as 18 & 19 year olds are teenagers. Same goes for the test method name.

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/EligibilityOverrideDTO.java
adeel718: Its bit confusing because QualifyingBenfits are returned from DWP, but now we are also using this as a qualifying reason. Can we create a new enum as matt suggested but with just under_18 and not_set?

AmarkumarJagtap: will keep it as is for now and will refactor in future if it's more confusing.

api/src/test/java/uk/gov/dhsc/htbhf/claimant/ClaimantLifecycleIntegrationTests.java
```diff
@@ -456,6 +494,17 @@ private UUID applyForHealthyStartOverridingEligibilityForAPregnantWomanWithNoChi
         return applyForHealthyStart(newClaimDTO, NO_CHILDREN);
     }
 
+    private UUID applyForHealthyStartOverridingEligibilityForUnder18Pregnant(LocalDate expectedDeliveryDate)
+            throws JsonProcessingException, NotificationClientException {
+        NewClaimDTO newClaimDTO = aValidClaimDTOWithEligibilityOverrideForUnder18Pregnant(
+                expectedDeliveryDate,
+                EligibilityOutcome.CONFIRMED,
+                OVERRIDE_UNTIL_FIVE_YEARS);
```
adeel718: Change this to 4 weeks, closer to the actual due date

api/src/test/java/uk/gov/dhsc/htbhf/claimant/testsupport/TestConstants.java
```diff
@@ -29,6 +29,7 @@
     public static final String TEST_EXCEPTION_MESSAGE = "test exception";
     public static final RuntimeException TEST_EXCEPTION = new RuntimeException(TEST_EXCEPTION_MESSAGE);
     public static final LocalDate OVERRIDE_UNTIL_FIVE_YEARS = LocalDate.now().plusYears(5);
+    public static final LocalDate OVERRIDE_UNTIL_TWENTY_NINE_WEEKS = LocalDate.now().plusYears(29);
```
YetAnotherMatt: This uses `plusYears` instead of `plusWeeks`

api/src/main/java/uk/gov/dhsc/htbhf/claimant/model/EligibilityOverrideDTO.java
dwybourn: I really think the change should be done now, I see no reason to delay work for the future if several people have said it's confusing.

YetAnotherMatt:
@AmarkumarJagtap shall we close this PR for the time being?

PR #567, opened 2020-02-12
AFHS-1956 - modified nino,email and phone number to be optional. 

AmarkumarJagtap:
Nino is mandatory field if claim is being created without eligibility override set. if nino is not set and eligibility override is not passed then claim gets rejected being ineligible

### DIFF
--- a/api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/DetermineEntitlementNotificationHandler.java
+++ b/api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/DetermineEntitlementNotificationHandler.java
@@ -20,14 +20,14 @@ public class DetermineEntitlementNotificationHandler {
 
     private final MessageQueueClient messageQueueClient;
 
-    public void sendClaimNoLongerEligibleEmail(Claim claim) {
+    public void sendClaimNoLongerEligibleEmailIfPresent(Claim claim) {
         if (StringUtils.isNotEmpty(claim.getClaimant().getEmailAddress())) {
             MessagePayload messagePayload = buildClaimIsNoLongerEligibleNotificationEmailPayload(claim);
             messageQueueClient.sendMessage(messagePayload, SEND_EMAIL);
         }
     }
 
-    public void sendNoChildrenOnFeedClaimNoLongerEligibleEmail(Claim claim) {
+    public void sendNoChildrenOnFeedClaimNoLongerEligibleEmailIfPresent(Claim claim) {
         if (StringUtils.isNotEmpty(claim.getClaimant().getEmailAddress())) {
             MessagePayload messagePayload = buildNoChildrenOnFeedClaimIsNoLongerEligibleNotificationEmailPayload(claim);
             messageQueueClient.sendMessage(messagePayload, SEND_EMAIL);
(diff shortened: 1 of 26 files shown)


api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/DetermineEntitlementNotificationHandler.java
```diff
@@ -20,13 +21,17 @@
     private final MessageQueueClient messageQueueClient;
 
     public void sendClaimNoLongerEligibleEmail(Claim claim) {
-        MessagePayload messagePayload = buildClaimIsNoLongerEligibleNotificationEmailPayload(claim);
-        messageQueueClient.sendMessage(messagePayload, SEND_EMAIL);
+        if (StringUtils.isNotEmpty(claim.getClaimant().getEmailAddress())) {
```
adeel718: I was thinking, rather than doing this and checking for null/ empty in all places shall we make the type Optional? In that case you can just call .isPresent

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/DetermineEntitlementNotificationHandler.java
```diff
@@ -20,13 +21,17 @@
     private final MessageQueueClient messageQueueClient;
 
     public void sendClaimNoLongerEligibleEmail(Claim claim) {
```
adeel718: Probably best if we don't even call this method if email is not present. In which case you won't have to change any logic here

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/DetermineEntitlementNotificationHandler.java
```diff
@@ -20,13 +21,17 @@
...
-        MessagePayload messagePayload = buildClaimIsNoLongerEligibleNotificationEmailPayload(claim);
-        messageQueueClient.sendMessage(messagePayload, SEND_EMAIL);
+        if (StringUtils.isNotEmpty(claim.getClaimant().getEmailAddress())) {
+            MessagePayload messagePayload = buildClaimIsNoLongerEligibleNotificationEmailPayload(claim);
+            messageQueueClient.sendMessage(messagePayload, SEND_EMAIL);
+        }
     }
 
     public void sendNoChildrenOnFeedClaimNoLongerEligibleEmail(Claim claim) {
```
adeel718: Same, do this check level above and not call this method

api/src/main/java/uk/gov/dhsc/htbhf/claimant/communications/PaymentCycleNotificationHandler.java
```diff
@@ -45,9 +46,11 @@ public void sendNotificationEmailsForRestartedPayment(PaymentCycle paymentCycle)
     }
 
     private void sendNotificationEmail(PaymentCycle paymentCycle, EmailType emailType) {
-        EmailMessagePayload messagePayload = emailMessagePayloadFactory.buildEmailMessagePayload(paymentCycle, emailType);
-        messageQueueClient.sendMessage(messagePayload, MessageType.SEND_EMAIL);
-        handleUpcomingBirthdayEmails(paymentCycle);
+        if (StringUtils.isNotEmpty(paymentCycle.getClaim().getClaimant().getEmailAddress())) {
```
adeel718: Same, don't even call this method. check if email exists level above

PR #571, opened 2020-03-03
AFHS-1927 user id implementation that needs to be set in request header as a part of claim creation process.

AmarkumarJagtap:
and also removed auto auditing of claim repository. audits are stored only in case of claim creation process


api/src/main/java/uk/gov/dhsc/htbhf/claimant/controller/ClaimController.java
```diff
@@ -52,11 +52,12 @@
     @PostMapping
     @ApiOperation("Create a claim.")
     @ApiResponses({@ApiResponse(code = 400, message = "Bad request", response = ErrorResponse.class)})
-    public ResponseEntity<ClaimResultDTO> createClaim(@RequestBody @Valid @ApiParam("The claim to persist") NewClaimDTO newClaimDTO) {
+    public ResponseEntity<ClaimResultDTO> createClaim(@RequestBody @Valid @ApiParam("The claim to persist") NewClaimDTO newClaimDTO,
```
adeel718: Is user optional as there will be no user passed through the applicant web journey

adeel718 (changes requested):

adeel718 (approved):

